Issue #4, opened 2018-02-07
Support tracing of HTTP requests using HttpClient


ljacobsson:
Feature request.

Currently the SDK only supports out-going HTTP requests using the `HttpWebRequest` client. It would be useful having tracing capabilities for those using `System.Net.HttpClient`.

For now I've created a [custom MessageHandler](https://github.com/mhlabs/MhLabs.AWSXRayHttpClientHandler/tree/master) for this, but would be great to have this or something similar as part of the SDK.

Edit: I just noticed that this has already been requested in #3.

I will look at adding that as a pull request

yogiraj07:
Hi @ljacobsson ,
Thank you for the feature suggestion. We welcome your pull request.

Thanks,
Yogi

SergeyKanzhelev:
You can use [Diagnostics Source](https://github.com/dotnet/corefx/blob/b3033e6bd586fb7bdc2e1afdd1694e790272ce6e/src/System.Net.Http/src/HttpDiagnosticsGuide.md) to get information about outgoing http calls.

yogiraj07:
Hi @ljacobsson ,

We have included this feature in `2.2.0-beta` release.

Please refer release details : https://github.com/aws/aws-xray-sdk-dotnet/blob/master/CHANGELOG.md

I am closing this issue.

Thanks,
Yogi

PR #5, opened 2018-02-07
Adding wrapper for HttpClientHandler

ljacobsson:
This is referring to feature request #4 

This provides tracing capabilities to `HttpClient` by passing in a custom `HttpClientHandler`. As part of this I pulled out the actual request/response processing into a utility class `RequestUtil`.

I placed this wrapper in project `Amazon.XRay.Recorder.Handlers.System.Net` although HttpClient is actually in the System.Net.Http namespace. I think it makes sense for it to live in the same project as `HttpWebRequestTracingExtension`, but please let me know if I should change the namespace to `Amazon.XRay.Recorder.Handlers.System.Net.Http`


yogiraj07:
Hi @ljacobsson,
Thank you for submitting pull requests. We would review it soon and also explore other options, if any for `HttpClient`.

Thanks,
Yogi

Issue #21, opened 2018-03-22
The XRayPipelineHandler will throw a null reference exception if the client does not have a RegionEndpoint assigned.


bittercoder:
The XRayPipelineHandler will throw a null reference exception if the client does not have a RegionEndpoint assigned.

https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs#L300

```
subsegment.Aws["region"] = client.RegionEndpoint.SystemName;
```

If for some reason you have an AWS service where you are explicitly setting the clients `ServiceURL` then the region endpoint will be null - in these cases this will cause the XRay SDK to blow up.

A good example of this is if your developing against local dev copies of AWS services (e.g. when employing local stack for local development).

It appears the AWS sub segment does not have anywhere to put the service URL for a request - but it does at least allow the region to be omitted if unavailable, so it wuold be great to change the code to be:

```
if (client.RegionEndpoint != null) {
    subsegment.Aws["region"] = client.RegionEndpoint.SystemName;
}
```

yogiraj07:
Hi @bittercoder ,

Thank you for raising this issue and submitting the pull request. You can leverage annotations and metadata feature for the service URL. Please refer : https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-dotnet-segment.html

Thanks,
Yogi

bittercoder:
@yogiraj07 so if I was to add service Url to the annotations for the segment in the AWS SDK handler - is that behaviour you would be happy to accept as a pull request?

yogiraj07:
Hello @bittercoder ,
We have made `2.2.1-beta` release. : [Changelog](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/CHANGELOG.md), which includes your PR. Closing this issue. Please feel free to reopen if you have any questions.

Thanks,
Yogi

PR #22, opened 2018-03-22
Update AWS SDK Tracing handling to handle RegionEndpoint being null

bittercoder:
Fix for issue https://github.com/aws/aws-xray-sdk-dotnet/issues/21

### DIFF
--- a/sdk/src/Handlers/AwsSdk/AWSSdkTracingHandler.cs
+++ b/sdk/src/Handlers/AwsSdk/AWSSdkTracingHandler.cs
@@ -301,7 +301,7 @@ namespace Amazon.XRay.Recorder.Handlers.AwsSdk
             var serviceName = RemoveAmazonPrefixFromServiceName(args.ServiceName);
             var operation = RemoveSuffix(args.Request.GetType().Name, "Request");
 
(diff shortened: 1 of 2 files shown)


sdk/src/Handlers/AwsSdk/AWSSdkTracingHandler.cs
```diff
@@ -301,7 +301,7 @@ private void AfterResponseEventHandler(object sender, ResponseEventArgs e)
             var serviceName = RemoveAmazonPrefixFromServiceName(args.ServiceName);
             var operation = RemoveSuffix(args.Request.GetType().Name, "Request");
 
-            subsegment.Aws["region"] = client.Config.RegionEndpoint.SystemName;
+            subsegment.Aws["region"] = client.Config.RegionEndpoint?.SystemName;
```
normj: If you have a ServiceUrl and not a RegionEndpoint then you call Amazon.Utils.AWSSDKUtils.DetermineRegion() passing in the service url to get the region for it. That way we could keep the feature of tracing the region.

bittercoder: Thanks for the suggestion @normj I will do a small PR to pick this up from the region, sounds reasonable.

Issue #19, opened 2018-03-22
Support specifying AWS_XRAY_DAEMON_ADDRESS as hostname:port as well as ip:port


bittercoder:
Feature request:

Having to specify AWS_XRAY_DAEMON_ADDRESS as ip:port is often no ideal if the IP address is somewhat dynamic - it would be great if hostname:port as also supported.

yogiraj07:
Hi @bittercoder ,

Thank you for the feedback. Can you please let us know your use case. I would take this as a feature request. 

Best,
Yogi

bittercoder:
One of the simple ones is if you have x-ray daemon as a side car linked container, I would rather specify the address as "xray-daemon:2000"

But more generally I just don't want to have to specify the IP address or deal with having to role containers if the address of the x-ray daemon changes for some reason.

So I guess I'm interested in know just have hostname:port work, but ensuring that once resolve the address lookup is only cached for say a minute before being resolved again.

yogiraj07:
Hi @bittercoder ,

Thanks for letting us know the use case. We take this as a feature request. Please feel free to submit a pull request, we would be happy to review it.

Best,
Yogi

emmekappa:
I have the same issue here

emmekappa:
I'm working on a PR here https://github.com/emmekappa/aws-xray-sdk-dotnet/tree/hostname_support

bittercoder:
Great to hear your going to work on a improvement for this @emmekappa 👍 

I've done something similar to this in the past which may or may not be useful to you:

```
public class HostEndpoint
{
	private static readonly Logger _logger = Logger.GetLogger(typeof(HostEndpoint));
	private readonly string _host;
	private readonly int _port;
	private readonly object _refreshLock = new object();
	private Stopwatch _age;
	private IPEndPoint _endpoint;

	public HostEndpoint(string host, int port)
	{
		_host = host;
		_port = port;
	}

	public TimeSpan RefreshInterval { get; set; } = TimeSpan.FromMinutes(1);

	public bool CacheNegativeResponses { get; set; } = true;

	private bool NeedRefresh => _age == null || _age.Elapsed > RefreshInterval;

	public bool TryGetEndpoint(out IPEndPoint endpoint)
	{
		if (NeedRefresh) Refresh();

		endpoint = _endpoint;
		return endpoint != null;
	}

	public static bool TryParse(string entry, out HostEndpoint hostEndpoint)
	{
[...]

PR #28, opened 2018-04-16
Hostname support

emmekappa:
Fixing issue https://github.com/aws/aws-xray-sdk-dotnet/issues/19 by supporting hostname as well as the IP address as XRay Daemon Address.


yogiraj07:
Hi @emmekappa ,
Thank you for submitting pull request. We will review it at the earliest.

Best,
Yogi

sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs
```diff
@@ -149,20 +151,34 @@ protected virtual void Dispose(bool disposing)
...
             {
-                daemonEndPoint = new IPEndPoint(_defaultDaemonAddress, _defaultDaemonPort);
-                _logger.InfoFormat("The given daemonAddress ({0}) is invalid, using default daemon address {1}:{2}.", daemonAddress, daemonEndPoint.Address.ToString(), daemonEndPoint.Port);
+                EndPoint = daemonEndPoint;
+            }
+            else if (HostEndpoint.TryParse(daemonAddress, out var hostEndpoint))
+            {
+                try
+                {
+                    EndPoint = hostEndpoint.GetIPEndPoint();
```
yogiraj07: Hi @emmekappa ,
Thank you for the PR.
Instead of getting a `IPEndPoint` instance for a hostname, should we just use `hostEndPoint `in the [_udpClient.Send()](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs#L77).

We do have inbuilt method provided by .NET : [Link](https://msdn.microsoft.com/en-us/library/acf44a1a(v=vs.110).aspx) on UDP client object, which accepts hostname and port.

For example the `send()` call would look like : `_udpClient.Send(data, data.Length, hostEndPoint.Host, hostEndPoint.Port);`
This will  eliminate the need of DNS look up and rely on underlying implementation for IP endpoint resolver.

Please let me know your thoughts.

yogiraj07: Hi @emmekappa ,
Would you like to provide your thoughts on the comment. Also, have you got chance on DNS caching implementation as per the thread : aws/aws-xray-sdk-dotnet#19 ?

Best,
Yogi

emmekappa: Hi @yogiraj07, sorry but I have had zero-time during the last weeks to work on this. 
We found too much issues by using this library and more in general the X-Ray service. I totally understand your points both on your feedback on my PR and the DNS caching issue but unfortunately I don't think I'll have time to work on this in a short period.

By the way I hope that my feedback could help you guys to improve the library, because I think that this could be a nice service for .NET developers.

yogiraj07:
Hi @emmekappa ,
Closing in favor of PR #38. Thank you for initial work.

Best,
Yogi

PR #35, opened 2018-06-17
Adding support for context missing option from IConfiguration

### LINKED ISSUE #33: Allow AWS_XRAY_CONTEXT_MISSING to be configured via IConfiguration
I've just started trying out this library today. One thing I found annoying with the default for AWS_XRAY_CONTEXT_MISSING to cause a runtime error. I'd like to deploy this onto existing production apps to see if it proves useful, but it's slightly riskier since if I miss any places where SDK calls, for example, are made outside of a normal ASP.NET Core request flow, it throws.

I realise I can set the AWS_XRAY_CONTEXT_MISSING environment variable to control this, but that adds work to deployments and if missed could cause the app to fail, purely over tracing needs. I'd much prefer a silent failure as the default here.

An alternative (and I've not dug into the code deeply yet) is that it would be nice to control this setting from the settings available in appSettings.json so I can set the default to logging only which I know will be deployed with the app. We can then override with an environment variable if we prefer.

stevejgordon:
Fixes #33 

Adds support for setting a UseRuntimeErrors option via .NET Core IConfiguration. When available the value from configuration will be applied. Otherwise, this defaults to true to ensure existing behaviour is respected.

If set, the value from AWS_XRAY_CONTEXT_MISSING will override the value from IConfiguration.

### DIFF
--- a/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
+++ b/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
@@ -84,7 +84,7 @@ namespace Amazon.XRay.Recorder.Core.Internal.Utils
         public bool IsXRayTracingDisabled { get => _isXRayTracingDisabled; set => _isXRayTracingDisabled = value; }
 
         /// <summary>
-        /// Determines whether runtime errors should occur when recording of data is attempted and a tracing header is not available, either true or false.
+        /// For missing Segments/Subsegments, if set to true, runtime exception is thrown, if set to false, runtime exceptions are avoided and logged.
         /// </summary>
         public bool UseRuntimeErrors { get => _useRuntimeErrors; set => _useRuntimeErrors = value; }
     }
--- a/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
+++ b/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
@@ -21,10 +21,11 @@ namespace Amazon.XRay.Recorder.Core.Internal.Utils
     /// </summary>
     public class XRayOptions
     {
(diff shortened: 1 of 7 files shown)


sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
```diff
@@ -63,9 +78,14 @@ public XRayOptions(string pluginSetting, string samplingRuleManifest, string aws
...
-        /// <summary>
-        /// Tracing disabled value, either true or false.
-        /// </summary>
+        /// <summary>
+        /// Tracing disabled value, either true or false.
+        /// </summary>
         public bool IsXRayTracingDisabled { get => _isXRayTracingDisabled; set => _isXRayTracingDisabled = value; }
+
+        /// <summary>
+        /// Determines whether runtime errors should occur when recording of data is attempted and a tracing header is not available, either true or false.
```
yogiraj07: Hi @stevejgordon ,
Can we modify this comment to `For missing Segments/Subsegments, if set to true, runtime exception is thrown, if set to false, runtime exceptions are avoided and logged.`
Reference:  "Environment Variables" section - https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-dotnet-configuration.html

stevejgordon:
@yogiraj07 Sorry for the delayed response! Thanks for the review. I'll look to make the update to the comment ASAP.

stevejgordon:
@yogiraj07 Sorry for the delay. Updated as requested.

yogiraj07:
Hi @stevejgordon ,
Thank you for your contribution. I have merged your PR.

Best,
Yogi

PR #38, opened 2018-08-17
Another go at hostname support

MrLarssonJr:
*Issue #19 

*Description of changes:*

Saw PR #28 discussing hostname support. This is my attempt at it.

Advantages of this design is that caches the resolved IP for the hostname.


yogiraj07:
Hi @MrLarssonJr ,
Thank you for submitting PR. We appreciate your efforts. I will review it at the earliest and keep you posted.

Thanks,
Yogi

DarrelH:
Hi @yogiraj07 - was curious if you might have a timeline? I'm hoping to avoid pushing workarounds into production

DarrelH:
Hi @yogiraj07 - was curious if you might have a timeline? I'm hoping to avoid pushing workarounds into production and @MrLarssonJr 's PR seems to work quite nicely

MrLarssonJr:
I want to note for the sake of transparency that this PR is largely based upon PR #28 that @emmekappa did. The addition of HostEndPoint in particular.

yogiraj07:
Hi @DarrelH , @MrLarssonJr ,

We are planning for a release soon. I would review the PR this week. Depending on code review completion and release date, we will either include the PR in the upcoming release else we will fast follow up and include this feature.

We appreciate your patience.

Thanks,
Yogi

emmekappa:
@MrLarssonJr thanks for the mention ;) I hope to see this feature available soon in the official SDK.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,102 @@
...
+					}
+					else
+					{
+						_logger.InfoFormat("IP cache invaild: DNS responded with zero IP addresses for {1}. Falling back to cached IP, e.g. {0}.", _ipCache, Host);
+					}
+				}
+				//Error catching for DNS resolve
+				catch(ArgumentNullException)
+				{
+					_logger.InfoFormat("IP cache invaild: failed to resolve DNS due to host being null. Falling back to cached IP, e.g. {0}.", _ipCache);
```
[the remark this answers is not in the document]
MrLarssonJr: Oh, nice catch 👍

MrLarssonJr:
I'll look into this this weekend.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,102 @@
...
+			{
+				return DateTime.Now.Subtract(lastTimestamp).TotalSeconds < CACHE_TTL;
+			}
+			else
+			{
+				return false;
+			}
+		}
+
+		public IPEndPoint GetIPEndPoint()
```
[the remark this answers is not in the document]
MrLarssonJr: 7572b7b4d333bfb6fcc408c01fdc7fa4bcdb50c3 addresses two threads concurrently updating the cache. That is the only write operation on the cache.

If two threads tries to update the cache concurrently only one will do so. The other will immediately return, opening up for the possibility that it will return a dirty cache. This is however not a problem due to we can't ever really guarantee the cache to not be dirty because the cached IP can become dirty before the TTL run out and will allow the critical path will still be speedy and not waiting on the DNS resolution of the first thread to complete.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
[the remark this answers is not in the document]
MrLarssonJr: Fixed in https://github.com/aws/aws-xray-sdk-dotnet/pull/38/commits/3c01af6fcf836ac660eb66ef84cd33d49d0a8366

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,102 @@
...
+				catch(ArgumentException)
+				{
+					_logger.InfoFormat("IP cache invaild: failed to update cache due to {0} not being a vaild IP.", Host);
+				}
+			}
+
+			return _ipCache;
+		}
+
+		public static bool TryParse(string entry, out HostEndPoint hostEndpoint)
```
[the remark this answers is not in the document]
MrLarssonJr: Shifted in https://github.com/aws/aws-xray-sdk-dotnet/pull/38/commits/eb52795419237021f5402d61294f829ff3b92b73

MrLarssonJr:
Hi sorry I was a bit late with this. School got in the way.

I've tried to merge and move stuff now. I still have to provide update the host input parsing in: https://github.com/aws/aws-xray-sdk-dotnet/blob/f8edd5f1688fe34bbc5ab12f491d82c600377852/sdk/src/Core/Internal/Utils/IPEndPointExtension.cs#L99

Unit tests and documentation also needs to be done.

I no longer have access to the sample app I had when I created this PR. I might be able to create a new one though.

yogiraj07:
Hi @MrLarssonJr,
Would you like to provide an update.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,133 @@
...
+			//If we fail with timeout = 0 another thread is already updating the cache and we don't need to it as well
+			bool entered = cacheLock.TryEnterReadLock(0);
+
+			if (!entered)
+			{
+				return;
+			}
+			
+			try
+			{
```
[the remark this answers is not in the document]
MrLarssonJr: Correct, my bad.

MrLarssonJr:
Hi, sorry for the delay.

The three most recent commits should address the points you raised in the review by @yogiraj07.

MrLarssonJr:
I think the host input parsing is okay and does not need any further adjustments. It should allow address on the following forms:
1. <hostname or address>:<port>
2. tcp:<hostname or address>:<port> udp:<hostname or address>:<port>

sdk/src/Core/Internal/Utils/DaemonConfig.cs
```diff
@@ -48,29 +48,44 @@ public class DaemonConfig
...
         /// <summary>
         /// Gets or sets TCP endpoint.
         /// </summary>
-        public IPEndPoint TCPEndpoint { get; set; }
+        public EndPoint TCPEndpoint { get; set; }
+        
+        /// <summary>
+        /// Gets IP for UDP endpoint.
+        /// </summary>
+        public IPEndPoint UDP_IP_Endpoint {
```
yogiraj07: Instead of adding new property for UDP and TCP, we can have internal variables `_udpEndpoint` and `_tcpEndpoint` of type `Endpoint`.
As a result, the original method `public IPEndPoint UDPEndpoint { get; set; }` will look like `public IPEndPoint UDPEndpoint { 
     get {_udpEndpoint.GetIPEndpoint()}
 }`

I think when users create instance of `DaemonConfig` the only public property should be `UDPEndpoint` and `TCPEndpoint` of type `IPEndPoint`. The getters on these should take care of returning correct IP Address.

The additional properties `TCP_IP_Endpoint` and `UDP_IP_Endpoint` may confuse the users. Please let me know your thoughts.

sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs
```diff
@@ -47,15 +48,17 @@ public UdpSegmentEmitter() : this(new JsonSegmentMarshaller())
...
+            _endPoint = daemonEndPoint.UDP_IP_Endpoint;
         }
 
+        private IPEndPoint _endPoint;
+            
         /// <summary>
         /// Gets the end point to daemon.
         /// </summary>
-        public IPEndPoint EndPoint { get; private set; }
+        public IPEndPoint EndPoint => _endPoint;
```
yogiraj07: We can have the original implementation : ` public IPEndPoint EndPoint { get; private set; }`

sdk/src/Core/Internal/Utils/EndPoint.cs
```diff
@@ -0,0 +1,43 @@
+using System.Net;
```
yogiraj07: Lets add the copyright documentation to all new files.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,151 @@
...
+using System.Net.Sockets;
+using System.Threading;
+using Amazon.Runtime.Internal.Util;
+
+namespace Amazon.XRay.Recorder.Core.Internal.Utils
+{
+	/// <summary>
+	/// Represents a endpoint on some network.
+	/// The represented endpoint is identified by a hostname.
+	///
```
yogiraj07: Trim space.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,151 @@
...
+		/// Get the hostname that identifies the endpoint.
+		/// </summary>
+		public string Host { get; }
+		/// <summary>
+		/// Get the port of the endpoint.
+		/// </summary>
+		public int Port { get; }
+
+
+		private bool isIPCacheValid()
```
yogiraj07: The SDK has all methods starting with uppercase. Lets maintain the consistency for all methods in this PR.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
MrLarssonJr: Do you refer to the leading space?

sdk/src/Core/Internal/Utils/DaemonConfig.cs
MrLarssonJr: I agree that the interface might be a bit clunky. The drawback of ```public IPEndPoint UDPEndpoint { get {_udpEndpoint.GetIPEndpoint()} }```is that the endpoint property becomes immutable from an outside perspective. But that might be okay?

sdk/src/Core/Internal/Utils/HostEndPoint.cs
MrLarssonJr: Addressed in 7cc3c2f

sdk/src/Core/Internal/Utils/EndPoint.cs
MrLarssonJr: Addressed in 06a1695

sdk/src/Core/Internal/Utils/EndPoint.cs
```diff
@@ -0,0 +1,43 @@
...
+        private HostEndPoint _h;
+        private IPEndPoint _i;
+        private bool _isHost;
+
+        /// <summary>
+        /// Create an EndPoint representing a HostEndPoint.
+        /// </summary>
+        /// <param name="hostEndPoint">the host endpoint to represent.</param>
+        /// <returns></returns>
+        public static EndPoint of(HostEndPoint hostEndPoint)
```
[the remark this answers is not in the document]
MrLarssonJr: Addressed in 7092d8e

sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs
MrLarssonJr: Addressed in b90e5ee

MrLarssonJr:
Do you have a guide on how to set up my environment to get the tests to pass? I saw a few tests that for example relies on S3 or dynamoDB.

I could just write some tests for the new code, but I wouldn't be able to guarantee that the changes have affected other parts of the SDK. Or maybe you are able to run tests in the PR before merging and reporting if somethings wrong?

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,168 @@
...
+		/// <returns></returns>
+		public IPEndPoint GetIPEndPoint()
+		{
+			
+			if (!IsIPCacheValid())
+			{
+				UpdateCache();
+			}
+			
+			return _ipCache;
```
yogiraj07: Should we have a read lock here ? If one thread has write lock on `_ipCache`, having read lock will guarantee consistent value else invalid value will be returned for the current thread.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,168 @@
...
+		/// Get the port of the endpoint.
+		/// </summary>
+		public int Port { get; }
+
+
+		private bool IsIPCacheValid()
+		{
+			bool entered = cacheLock.TryEnterReadLock(0);
+
+			if (!entered)
```
yogiraj07: I think we should not check this condition. The reason : `GetIPEndPoint()` calls this method. So suppose, 1 thread return value false, and enters `UpdateCache()` code block with the write lock and at the same time another thread hits `IsIPCacheValid()`, the current logic of returning value `false` at this point  will result in calling `UpdateCache()` unnecessarily for the second time. `UpdateCache()`is expensive call since it has write lock with DNS resolution operation. As per convention, lets have read lock in try finally block, to avoid any kind of exception  and change read lock to `EnterReadLock`

sdk/test/UnitTests/IPEndPointExceptionTests.cs
```diff
@@ -60,21 +60,55 @@ public class IPEndPointExceptionTests
...
             IPEndPoint endpoint;
-            foreach (var ip in _validInputs)
+            foreach (var ip in _validIPs)
             {
                 Assert.IsTrue(IPEndPointExtension.TryParse(ip, out endpoint));
             }
         }
 
+        [TestMethod]
+        public void TestValidDomains()
```
yogiraj07: Nice basic test. Lets add few tests here : https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/test/UnitTests/DaemonConfigTests.cs

The file covers all major tests for IPs, please add for domain name. We need to test `UpdateCache()`. Let me know if you can cover this in unit tests.

yogiraj07:
Hi @MrLarssonJr ,


> Do you have a guide on how to set up my environment to get the tests to pass? I saw a few tests that for example relies on S3 or dynamoDB.
> 
> I could just write some tests for the new code, but I wouldn't be able to guarantee that the changes have affected other parts of the SDK. Or maybe you are able to run tests in the PR before merging and reporting if somethings wrong?

Hi @MrLarssonJr ,
Once you clone the repo and build SDK, the unit tests should run locally. You don't need any extra configuration. For integration tests, you would need to run [daemon](https://docs.aws.amazon.com/xray/latest/devguide/xray-daemon-local.html) locally in region `us-east-1`

Do you have any issue running unit tests locally ?

I would definitely help you with testing. Do you have any sample app for integration tests, I would like to check the `UpdateCache()` functionality with concurrent threads.

Thank you for your patience,
Yogi

yogiraj07:
Hi @MrLarssonJr ,
Can you please provide an update ?

Thanks,
Yogi

MrLarssonJr:
Hi,
Excuse me for being absent a while. I got caught up in exams.

I think I've figured out why I had problem with running the tests. It seems that some paths assumes Windows path separators, causing problem on my macOS system.
One example of this is in the file ```./sdk/test/UnitTests/AWSSdkTracingHandlerTests.cs```where ```private static String _path = @"JSONs\AWSRequestInfo.json";```occurs. I'll create an issue.

yogiraj07:
Hi @MrLarssonJr ,
I will take a look at the reported issue soon. Can you please expedite this PR, its been a while and we would like to include this feature at the earliest.

MrLarssonJr:
Hi, sorry I've been lagging a bit.

I submitted an PR which I believe fixes the issues of platform dependence in the test suite.

I shall try to get a few hours of work in the coming week, but term, and with it deadlines, is quickly coming to an end right now. But I definitely not dropping the ball, I am interested in seeing this getting up to spec and then merged.

MrLarssonJr:
I saw that #42 was merged into master now. Awesome!

sdk/src/Core/Internal/Utils/HostEndPoint.cs
MrLarssonJr: Did som rework on HostEndPoint so that the double update no longer occurs. Also included unit tests to check for that.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
MrLarssonJr: Addressed in the HostEndPoint rework.

sdk/src/Core/Internal/Utils/DaemonConfig.cs
MrLarssonJr: I think that 36c1c83 simplifies that interface as requested.

yogiraj07:
Hi @MrLarssonJr ,
Thanks for the update. Will review it soon.

sdk/src/Core/Internal/Utils/IPEndPointExtension.cs
```diff
@@ -95,6 +95,70 @@ public static bool TryParse(string input, out IPEndPoint endPoint)
...
+
+        /// <summary>
+        /// Tries to parse a string to <see cref="EndPoint"/>
+        /// </summary>
+        /// <param name="input">The input string.</param>
+        /// <param name="hostEndpoint">The parsed EndPoint</param>
+        /// <returns>true if <paramref name="input"/> converted successfully; otherwise, false.</returns>
+        public static bool TryParse(string input, out EndPoint endpoint)
+        {
+            if (TryParse(input, out IPEndPoint ipEndPoint))
```
yogiraj07: If `Hostname` is passed, this condition is false. However we get a log message `Failed to parse IPEndPoint because input is invalid.` because of this [check](https://github.com/aws/aws-xray-sdk-dotnet/blob/36c1c83e3d8fdf49763ffe15be84facd3d4cf45c/sdk/src/Core/Internal/Utils/IPEndPointExtension.cs#L54).
This log can be confusing for the customer who pass hostname. Instead, should we have helper method, say `IsIPAddress()` which validates whether its valid IP, if yes we call `(TryParse(input, out IPEndPoint ipEndPoint))` `else if (TryParse(input, out HostEndPoint hostEndPoint))`.

This approach helps in clean log messages.

sdk/test/UnitTests/IPEndPointExceptionTests.cs
yogiraj07: Any update on this?

With hostname support lets include tests [here](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/test/UnitTests/DaemonConfigTests.cs).

The combination of valid inputs are : 
```
hostname:3001
tcp:hostname:3000 udp:127.0.0.1:3001
tcp:hostname:3000 udp:hostname:3001
```

Lets have `hostname` = `localhost` and in assertion check whether IP is `127.0.0.1`. This will validate resolving logic.

sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs
```diff
@@ -47,7 +48,7 @@ public UdpSegmentEmitter() : this(new JsonSegmentMarshaller())
         private UdpSegmentEmitter(ISegmentMarshaller marshaller)
         {
             _marshaller = marshaller;
-            _udpClient = new UdpClient();
+            _udpClient = new UdpClient();
             DaemonConfig daemonEndPoint = DaemonConfig.GetEndPoint();
             EndPoint = daemonEndPoint.UDPEndpoint;
```
yogiraj07: In the current implementation, `EndPoint = daemonEndPoint.UDPEndpoint;`, whenever, hostname is refreshed, the udp emitter `Endpoint` is not refreshed since `get` on `daemonEndpoint` is called only once, here.

Instead, we need to store `DaemonConfig` instance as a private member of this class. 
`private DaemonConfig _daemonConfig`.
In constructor, it would be : `_daemonConfig = DaemonConfig` and same [here](https://github.com/aws/aws-xray-sdk-dotnet/blob/36c1c83e3d8fdf49763ffe15be84facd3d4cf45c/sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs#L147).

And this [point](https://github.com/aws/aws-xray-sdk-dotnet/blob/36c1c83e3d8fdf49763ffe15be84facd3d4cf45c/sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs#L71), the code change is ` _udpClient.Send(data, data.Length, _daemonConfig.UDPEndpoint);`.

At `Debug` level we can have this log `_logger.DebugFormat("UDP Segment emitter enpoint: " + _daemonConfig.UDPEndpoint);`
This ensures, UDP emitter is updated in correct manner.

yogiraj07:
Hi @MrLarssonJr ,
Any updates ?

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,246 @@
...
+		/// <summary>
+		/// Get the hostname that identifies the endpoint.
+		/// </summary>
+		public string Host { get; }
+		/// <summary>
+		/// Get the port of the endpoint.
+		/// </summary>
+		public int Port { get; }
+		
+		public bool IsInitialised { get; private set; }
```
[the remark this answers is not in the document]
MrLarssonJr: Removed in f56763baf35f91483a4fbae24de83e5f1c9f3f9b.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,246 @@
...
+		/// This method is essentially a wrapper around <see cref="IPCacheIsValid"/> that acquires the required lock.
+		/// </summary>
+		/// <returns>true if the cache is valid, false otherwise.</returns>
+		private CacheState LockedIPCacheIsValid()
+		{
+			// If !entered => another thread holds a write lock, i.e. updating the cache
+			if (!cacheLock.TryEnterReadLock(0)) return CacheState.Updating;
+			try
+			{
+
```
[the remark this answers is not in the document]
MrLarssonJr: Fixed in c2e986da05d0920aee8e26a28617ee7db8819139.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,246 @@
...
+						_ipCache, Host);
+					return false;
+				}
+				// Upgrade our read lock to write mode
+				cacheLock.EnterWriteLock();
+				try
+				{
+					_timestampOfLastIPCacheUpdate = DateTime.Now;
+					_ipCache = new IPEndPoint(newIP, Port);
+					IsInitialised = true;
```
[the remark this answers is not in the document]
MrLarssonJr: Fixed in 8908ebea2bfde3c15b546f8e038fde92f4559e96.

sdk/src/Core/Internal/Utils/HostEndPoint.cs
```diff
@@ -0,0 +1,246 @@
...
+				catch (ArgumentOutOfRangeException)
+				{
+					_logger.InfoFormat("IP cache invalid: either the port {0} or IP address {1} is out of range. Cached IP address not updated!", Port, newIP);
+				}
+				finally
+				{
+					// Downgrade back to read mode
+					cacheLock.ExitWriteLock();
+				}
+				_logger.InfoFormat("IP cache invalid: updated ip cache for {0} to {1}.", Host, newIP);
```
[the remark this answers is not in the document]
MrLarssonJr: Fixed in 8908ebea2bfde3c15b546f8e038fde92f4559e96.

sdk/src/Core/Internal/Utils/IPEndPointExtension.cs
MrLarssonJr: Addressed in commit 4e896fcf522dedaf5a0006fcbbe39c8217e25d2f.

sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs
MrLarssonJr: Sorted in d95c7b0c698aac7808595a2c200fc1dc78561af4.

MrLarssonJr:
Hi @yogiraj07,
Sorry for the absence. Addressed a few of your point tonight. If I'm keeping track correctly, what's left is:

- rounding out tests, including for IPEndPointExtension.
- rebase onto master, and
- your point 2, about service connector

Could you please confirm?

Thank you for your patience.

sdk/src/Core/Internal/Utils/IPEndPointExtension.cs
```diff
@@ -143,21 +147,27 @@ public static bool TryParse(string input, out HostEndPoint hostEndpoint)
...
+                {
+                    endpoint = EndPoint.Of(ipEndPoint);
+                    return true;
+                }
             }
             else
             {
-                endpoint = null;
-                return false;
+                _logger.DebugFormat("Determined that {0} is not an IP.", input);
```
yogiraj07: can we modify this message to ` _logger.DebugFormat("Determined that {0} is not an IP address, would try to parse it as hostname.", input);`

Absence of this may confuse the user.

MrLarssonJr: Addressed in 0de9693.

MrLarssonJr:
I think I've rounded out the tests now.

sdk/test/UnitTests/IPEndPointExceptionTests.cs
yogiraj07: Sure, `TestMultiThreadedUpdateSpacing` works.

MrLarssonJr:
I've added `RefreshEndPoint()` to `ServiceConnector` now with needed instance members and corresponding simplifications. It relies on comparing service URLs instead of a `GetTCPEndpoint(out bool isUpdated)` method because I assume that the `DaemonConfig` instance might be used other places, meaning that `ServiceConnector` might not be the caller that get's the update notification.

A worry that I have with the locks I put in place in `GetSamplingTargets` and `GetSamplingRules` are thet they are on some critical path and those locks introduces intolerable delays. What do you think?

yogiraj07:
> I've added `RefreshEndPoint()` to `ServiceConnector` now with needed instance members and corresponding simplifications. It relies on comparing service URLs instead of a `GetTCPEndpoint(out bool isUpdated)` method because I assume that the `DaemonConfig` instance might be used other places, meaning that `ServiceConnector` might not be the caller that get's the update notification.
> 
> A worry that I have with the locks I put in place in `GetSamplingTargets` and `GetSamplingRules` are thet they are on some critical path and those locks introduces intolerable delays. @yogiraj07, what do you think?

I think this is a valid point. +1 for optimization. I would perform local testing soon. I appreciate your efforts.

MrLarssonJr:
Thanks!

And just to confirm; are you saying you will perform local testing? Or are you suggesting I should perform some? It would be great if you could do that!

yogiraj07:
> Thanks!
> 
> And just to confirm; are you saying you will perform local testing? Or are you suggesting I should perform some? It would be great if you could do that!

I would be doing local testing.

MrLarssonJr:
Hi, I looked into rebasing my fork onto master today and realised that I've in the past merged the master at a few points instead of performing rebases. Is it desired to have a super clean history, or is it ok if I use merges?

yogiraj07:
@MrLarssonJr , I would prefer clean history, unless merging is really required. This helps in clean commits and easy for rest developers for references. Let me know, once you have finally rebased this PR. I would perform some testing.

Best,
Yogi

MrLarssonJr:
Is it high priority for all of my commits to be present or is it acceptable if I squash my forks history?

MrLarssonJr:
Never mind, I managed to untangle the history and rebase the branch on top of master.

rui-ktei:
Can I ask when will this change be released? I'm keen to use xray with ECS but without this I can use it properly.

yogiraj07:
Hi @MrLarssonJr ,
Thank you for working with me. I have merged the PR.
@rui-ktei , we have planned release soon and this feature is included.

Thanks for the patience.

rui-ktei:
@yogiraj07 thank you!

MrLarssonJr:
Hi @yogiraj07,
Awesome! Thank you for you're time and patience, it has been pleasure working with you.

PR #40, opened 2018-09-28
Prevent X-Amzn-Trace-Id from Being Added More Than Once

chrisoverzero:
Under some circumstances, the chain of calls through
`DelegatingHandler` instances can reach a particular handler more than
once. When that handler is `HttpClientXRayTracingHandler`, it produces
an X-Amzn-Trace-Id header that cannot be parsed because it is
_multiple_ headers. (As represented in a string by separating them
with ", ".)

This patch checks whether X-Amzn-Trace-Id has already been added, and
skips adding another if it already has been.

*Issue #, if available:*

N/A

*Description of changes:*

Adds X-Amzn-Trace-Id at most once.

### DIFF
--- a/sdk/src/Handlers/System.Net/Utils/RequestUtil.cs
+++ b/sdk/src/Handlers/System.Net/Utils/RequestUtil.cs
@@ -24,15 +24,12 @@ namespace Amazon.XRay.Recorder.Handlers.System.Net.Utils
         /// <param name="request">An instance of <see cref="HttpRequestMessage"/></param>
         internal static void ProcessRequest(HttpRequestMessage request)
         {
-            ProcessRequest(request.RequestUri, request.Method.Method, AddHeader);
+            ProcessRequest(request.RequestUri, request.Method.Method, AddOrReplaceHeader);
 
-            void AddHeader(string header)
+            void AddOrReplaceHeader(string header)
             {
-                // If we've already added a trace header, don't add another one.
-                if (!request.Headers.Contains(TraceHeader.HeaderKey))
-                {
-                    request.Headers.Add(TraceHeader.HeaderKey, header);
-                }
+                request.Headers.Remove(TraceHeader.HeaderKey);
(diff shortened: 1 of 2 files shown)


yogiraj07:
Hi @chrisoverzero ,
Thank you for the PR. Can you provide with sample app/ steps to reproduce this issue ? If you see any exception, please provide stack trace. 
Since `DelegatingHandler` is called multiple times, this means [BeginSubsegment](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/System.Net/Utils/RequestUtil.cs#L52) is overridden in each call for the same request.

At any point , latest [header](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/System.Net/Utils/RequestUtil.cs#L63) value should be put in the request header, so that, if a downstream service is instrumented with X-Ray, the Segment/Subsegment generated by the the downstream service, gets connected with the calling service. 


Best,
Yogi

chrisoverzero:
I don’t have a reducible case at the moment, I'm afraid. The scenario is that we have a delegating handler that handles 401 responses from web services by retrieving an auth token. The stack of delegating handlers reaches the X-Ray handler twice when we have a cache miss. If you really want a sample, I can synthesize one, but I think the problem is rather clear when considering how the `HttpRequestHeaders.Add` method works. 

The header ends up with a value that looks like this when stringified: "Root=xyzxyzxyz; Shared=1, Root=xyzxyz; Shared=1". When server-side X-Ray attempts to parse this header, it splits it on semicolon, then splits on equals, then crashes when calling `ToDictionary` because the key "Shared" is duplicated. 

It sounds as though I got the fix completely backwards. Is that what you're telling me? The desired semantics are that if the "X-Amzn-Trace-Id" header is already present, we want to _replace_ it entirely. I can get another push tomorrow by noon Eastern.

chrisoverzero:
Great! Thank you so much.

PR #49, opened 2018-12-12
Add command text to SQL command

jaredcnance:
*Issue #, if available:* None

*Description of changes:* Adds the parameterized SQL command to the metadata. I tried adding it using the `AddSqlInformation` API, but it isn't available in the console/UI so I suspect there is a requisite backend change to handle this. I don't really expect this to get merged, but hope to either get some advice on how this _should_ be done today, or start a discussion on how it should be done in the future.

Example content of `CommandText` (generated by EntityFrameworkCore):

```
SELECT TOP(1) [e].[Id], [e].[Name]\nFROM [users] AS [e]\nWHERE [e].[Id] = @__get_User_0
```

It would be nice to strip the newlines out or pretty print them. That probably shouldn't be done application side since it could result in lots of string allocations, but it might make sense on the console side of things, especially if it receives special treatment as SQL command text.

### DIFF
--- a/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
+++ b/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
@@ -21,12 +21,6 @@ namespace Amazon.XRay.Recorder.Core.Internal.Utils
     /// </summary>
     public class XRayOptions
     {
-        private string _pluginSetting;
-        private string _samplingRuleManifest;
-        private string _awsServiceHandlerManifest;
-        private bool _isXRayTracingDisabled;
-        private bool _useRuntimeErrors = true;
-
         /// <summary>
         /// Default constructor.
         /// </summary>
@@ -54,38 +48,47 @@ namespace Amazon.XRay.Recorder.Core.Internal.Utils
         /// <param name="awsServiceHandlerManifest">AWS Service manifest file path.</param>
         /// <param name="isXRayTracingDisabled">Tracing disabled value, either true or false.</param>
         /// <param name="useRuntimeErrors">Should errors be thrown at runtime if segment not started, either true or false.</param>
-        public XRayOptions(string pluginSetting, string samplingRuleManifest, string awsServiceHandlerManifest, bool isXRayTracingDisabled, bool useRuntimeErrors)
+        public XRayOptions(string pluginSetting, string samplingRuleManifest, string awsServiceHandlerManifest, bool isXRayTracingDisabled, bool useRuntimeErrors, bool collectSqlQueries = false)
         {
             PluginSetting = pluginSetting;
             SamplingRuleManifest = samplingRuleManifest;
             AwsServiceHandlerManifest = awsServiceHandlerManifest;
             IsXRayTracingDisabled = isXRayTracingDisabled;
             UseRuntimeErrors = useRuntimeErrors;
+            CollectSqlQueries = collectSqlQueries;
         }
 
         /// <summary>
         /// Plugin setting.
         /// </summary>
-        public string PluginSetting { get => _pluginSetting; set => _pluginSetting = value; }
+        public string PluginSetting { get; set; }
 
         /// <summary>
         /// Sampling rule file path.
         /// </summary>
-        public string SamplingRuleManifest { get => _samplingRuleManifest; set => _samplingRuleManifest = value; }
+        public string SamplingRuleManifest { get; set; }
 
         /// <summary>
         /// AWS Service manifest file path.
         /// </summary>
-        public string AwsServiceHandlerManifest { get => _awsServiceHandlerManifest; set => _awsServiceHandlerManifest = value; }
+        public string AwsServiceHandlerManifest { get; set; }
 
         /// <summary>
         /// Tracing disabled value, either true or false.
         /// </summary>
-        public bool IsXRayTracingDisabled { get => _isXRayTracingDisabled; set => _isXRayTracingDisabled = value; }
+        public bool IsXRayTracingDisabled { get; set; }
 
         /// <summary>
         /// For missing Segments/Subsegments, if set to true, runtime exception is thrown, if set to false, runtime exceptions are avoided and logged.
         /// </summary>
-        public bool UseRuntimeErrors { get => _useRuntimeErrors; set => _useRuntimeErrors = value; }
+        public bool UseRuntimeErrors { get; set; } = true;
+
+        /// <summary>
+        /// Include the <see cref="TraceableSqlCommand.CommandText" /> in the sanitized_query.
+        /// Parameterized values will appear in their tokenized form and will not be expanded.
+        /// You should not enable this flag if you are not including sensitive information as clear text.
+        /// This flag can also be overridden for each <see cref="TraceableSqlCommand" /> instance individually.
+        /// </summary>
+        public bool CollectSqlQueries { get; set; }
     }
 }
--- a/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
+++ b/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
@@ -49,10 +49,10 @@ namespace Amazon.XRay.Recorder.Core.Internal.Utils
         /// <param name="isXRayTracingDisabled">Tracing disabled value, either true or false.</param>
         /// <param name="useRuntimeErrors">Should errors be thrown at runtime if segment not started, either true or false.</param>
         /// <param name="collectSqlQueries">
-        /// Include the <see cref="TraceableSqlCommand.CommandText" /> in the sanitized_query section of 
+        /// Include the TraceableSqlCommand.CommandText in the sanitized_query section of 
         /// the SQL subsegment. Parameterized values will appear in their tokenized form and will not be expanded.
         /// You should not enable this flag if you are including sensitive information as clear text.
-        /// This flag can also be overridden for each <see cref="TraceableSqlCommand" /> instance individually.
+        /// This flag can also be overridden for each TraceableSqlCommand instance individually.
         /// See the official documentation on <a href="https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.parameters?view=netframework-4.7.2">SqlCommand.Parameters</a>
         /// </param>
         public XRayOptions(string pluginSetting, string samplingRuleManifest, string awsServiceHandlerManifest, bool isXRayTracingDisabled, bool useRuntimeErrors, bool collectSqlQueries = false)
--- a/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
+++ b/sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
@@ -48,6 +48,13 @@ namespace Amazon.XRay.Recorder.Core.Internal.Utils
         /// <param name="awsServiceHandlerManifest">AWS Service manifest file path.</param>
         /// <param name="isXRayTracingDisabled">Tracing disabled value, either true or false.</param>
         /// <param name="useRuntimeErrors">Should errors be thrown at runtime if segment not started, either true or false.</param>
+        /// <param name="collectSqlQueries">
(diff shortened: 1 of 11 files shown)


yogiraj07:
Hi @jaredcnance ,
AWS X-Ray records sanitized query in `sql` namespace under [sanitized_query](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html) field.
So the correct way to use is `recorder. AddSqlInformation("sanitized_query","query to be recorded");` instead of meta data. So can you please update this PR [here](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/SqlServer/TraceableSqlCommand.netstandard.cs#L454)

We will get our security team review it as soon as possible.

In the mean time could you confirm `CommandText` would be sufficient to record. The API reference [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.data.sqlclient.sqlcommand.commandtext?view=netframework-4.7.2) doesn't mention anything about stripping or recording user data.

Thank you for the contribution.

Best
Yogi

jaredcnance:
@yogiraj07 I've updated the PR to use `sanitized_query` instead. `CommandText` is the raw query less the paramterized values. In general, user information should be specified as query parameters. 

In the absence of an ORM you would write your queries like:

```c#
const string userId = "my-id";
using (var conn = new SqlConnection(connString))
{
    var cmd = new SqlCommand("Select * From users Where userId=@userId", conn);
    cmd.Parameters.Add(new SqlParameter("@userId", userId));

    // the command text will not contain the parameter values
    Assert.DoesNotContain(cmd.CommandText, userId);

    // send the query
    await conn.OpenAsync();
    using(var reader = await cmd.ExecuteReaderAsync())
        while(await reader.ReadAsync()) { /* ... */ }
}
```

This puts the burden on the developers but also avoids any magic in-process string replacement and unnecessary allocations. What are the other SDKs doing to strip out user data?

yogiraj07:
Hi @jaredcnance ,
`CollectSqlInformation` is a helper method. I think, making [Intercept()](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/SqlServer/TraceableSqlCommand.netstandard.cs#L420) and [InterceptAsync()](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/SqlServer/TraceableSqlCommand.netstandard.cs#L397) method protected is a good choice considering maintainability of the code and giving flexibility for the users to write their own implementation on the main method.

jaredcnance:
@yogiraj07 I agree that `Intercept`/`InterceptAsync` should likely be `protected virtual` but I would contend that `CollectSqlInformation` should also be overridable because of the targeted nature of the method and a known use case. I don't think adding this method to the public API surface area adds much risk or complexity considering the nature of the method. 

All I want to do is add data to the published payload. As an application developer, I shouldn't be concerned with beginning and ending subsegments and I shouldn't have to override 2 methods just to add the `sanitized_query`.

yogiraj07:
Hi @jaredcnance ,
I think adding a feature flag to opt-in for recording sanitized query, to all [TraceableSqlCommand](https://github.com/aws/aws-xray-sdk-dotnet/blob/c7f193649251a1d68221f8525b5e04f2f9ae6b9d/sdk/src/Handlers/SqlServer/TraceableSqlCommand.netstandard.cs#L60) constructors is a good idea. By default this flag will be false and no query will be recorded. This will help in conscious decision.

In addition to this, I think `Intercept/InterceptAsync` can be protected virtual, if the users want there own implementation of the interceptors. 

Can you please make the changes to both [.NET](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/SqlServer/TraceableSqlCommand.net45.cs) and [.NET Core](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/SqlServer/TraceableSqlCommand.netstandard.cs) platform.

Thanks,
Yogi

yogiraj07:
Hi @jaredcnance ,
Would you like to provide some update? Or else I can work on this PR for its completion.

Please let me know.

Thanks,
Yogi

jaredcnance:
@yogiraj07 i have this completed already but it's my opinion that there need to be tests on this before it rolls out. The current design makes testing `TraceableSqlCommand` incredibly difficult for a few reasons. The first is that no abstraction has been created for the underlying `SqlCommand`. There are a couple issues here:

1. SqlCommand is `sealed`
2. We depend on return types that are specific to the `SqlCommand` type rather than the abstract `DbCommand` return types.

So, mocking/stubbing the sql operations is not easy without some ugly decisions (such as creating an `SqlCommandFactory` that returns a proxy to `SqlCommand`, so I decided to see if I could inherit `TraceableSqlCommand` and test the protected methods (`Intercept`)  directly. Unfortunately, this led me to the next issue:

3. `TraceableSqlCommand` has an implicit dependency on `AWSXRayRecorder.Instance`.`Instance` returns the concrete type `AWSXRayRecorder` rather than the interface `IAWSXRayRecorder`. It also depends on methods that don't exist on the interface. 

So, at a minimum we need to refactor the dependency on `AWSXRayRecorder` so that it depends on the interfaces. 

In addition to the tests, I think configuring the collection of SQL queries should be 2 fold:

1. Global option on `XRayOptions`
2. Local option in constructor `TraceableSqlCommand(bool collectSqlQueries = false)` that if provided overrides the global option.

Let me know what your thoughts are and I can finish this work up.

yogiraj07:
Hi @jaredcnance ,
I worked internally on your comments.
`AWSXRayRecorder.Instance` return type probably is not the problem, but `Intercept()` accessing [Connection](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/SqlServer/TraceableSqlCommand.netstandard.cs#L448).
With research on mocking this class and your analysis, I think going ahead with `SqlCommandFactory` may be the path forward. This will not be a breaking change.

I think, XRayOptions and constructor options, as 2 way opt-in way is okay. However, we can document it clearly the precedence of constructor value is over the  XRayOptions value. By default, in either case, the value will be `false`. 

We also need tests for [XRayOptions](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/test/UnitTests/TestXRayOptions.cs).

Once the PR is finalize, I will do a fast follow-up for NET45.

Sure, I have CI for the repo in the backlog and will prioritize it soon.

sdk/src/Handlers/SqlServer/TraceableSqlCommand.netstandard.cs
```diff
@@ -394,67 +426,10 @@ protected override DbDataReader ExecuteDbDataReader(CommandBehavior behavior)
...
-            {
-                recorder.AddException(e);
-                throw;
-            }
-            finally
-            {
-                recorder.EndSubsegment();
-            }
-        }
+        protected async Task<TResult> InterceptAsync<TResult>(Func<Task<TResult>> method)
```
yogiraj07: Also add `virtual` modifier?

sdk/test/UnitTests/DbCommandInterceptorTests.cs
```diff
@@ -0,0 +1,223 @@
...
+
+            // act
+            await interceptor.InterceptAsync(() => Task.FromResult(0), _command);
+
+            // assert
+            AssertNotCollected(recorder);
+            recorder.EndSegment();
+        }
+
+        private void AssertNotCollected(AWSXRayRecorder recorder)
```
yogiraj07: `AssertNotCollected` and `AssertCollected` should also assert rest parameters in `CollectSqlInformation()`.

jaredcnance: I’ve added them because there aren’t any other tests on this, but in general, I don’t think you should make assertions that lie outside the scope of the test. In an ideal world (this never happens), if something is broken you would have a single test fail that points to exactly what the issue is rather than lots of failing tests. And you don’t want to have to update a bunch of tests if you add a member to the payload. 🤷‍♂️

sdk/src/Core/Internal/Utils/AppSettings.netcore.cs
```diff
@@ -43,49 +37,58 @@ public XRayOptions()
...
         /// <summary>
         /// For missing Segments/Subsegments, if set to true, runtime exception is thrown, if set to false, runtime exceptions are avoided and logged.
         /// </summary>
-        public bool UseRuntimeErrors { get => _useRuntimeErrors; set => _useRuntimeErrors = value; }
+        public bool UseRuntimeErrors { get; set; } = true;
+
+        /// <summary>
+        /// Include the <see cref="TraceableSqlCommand.CommandText" /> in the sanitized_query.
+        /// Parameterized values will appear in their tokenized form and will not be expanded.
+        /// You should not enable this flag if you are not including sensitive information as clear text.
```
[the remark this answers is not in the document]
yogiraj07: Hey, I think we should rephrase this to `You should not enable this flag if you are including sensitive information as clear text.`

sdk/test/UnitTests/DbCommandInterceptorTests.cs
yogiraj07: @jaredcnance , agree. For the moment feel free to have a new test which checks rest of the parameters. Or else current version works.

yogiraj07:
Hi @jaredcnance ,
Apologies in delay in response.
I am performing a test to set "CollectSqlQueries : true" in appsettings.json.

This is my json 
```json
{
  "XRay": {
    "DisableXRayTracing": "false",
    "SamplingRuleManifest": "sampling-rules.json",
    "UseRuntimeErrors": "true",
    "CollectSqlQueries": "true"
  }
}

```

However, [ShouldCollectSqlText](https://github.com/aws/aws-xray-sdk-dotnet/blob/06ecc16d90a065228afb798e1b72703f886619c8/sdk/src/Handlers/SqlServer/DbCommandInterceptor.cs#L142) returns false.

The reason is, `CollectSqlQueries` key is not read from appsetting.json file. You would need to add key here: [ConfigurationExtensions](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/Internal/Utils/ConfigurationExtensions.cs).

jaredcnance:
Ah, I see. Instead of the `GetSetting` methods, I would have expected the `IConfiguration["XRay"]` block to be directly deserialized to `XRayOptions` e.g.:

```c#
services.Configure<XRayOptions>(Configuration.GetSection("XRay"));

// or

var options = new XRayOptions();
Configuration.GetSection("XRay").Bind("Weblog", config);
```

But, for now I can make the change to add the `CollectSqlQueries` key.

yogiraj07:
@jaredcnance , good point. For the moment you can also add [unit test](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/test/UnitTests/TestXRayOptions.cs) for the new configuration.

rtablada:
Any update on this @jaredcnance or @yogiraj07? What is missing to get this merged and a version published?

yogiraj07:
@jaredcnance , can you submit another revision with necessary changes requested? I have added CI, the current PR fails while building since the changes are only for NET Core. The CI will help you to make changes only for NET Core platform. If you are okay, I can make those changes on your behalf and expedite the implementation for NET 45 framework.

jaredcnance:
@yogiraj07 go ahead and commit on top of my branch, I don't think I will have cycles to work on this anytime soon.

sdk/src/Core/Internal/Utils/ConfigurationExtensions.cs
```diff
@@ -31,6 +31,7 @@ public static class XRayConfiguration
         private const string AWSServiceHandlerManifestKey = "AWSServiceHandlerManifest";
         private const string DisableXRayTracingKey = "DisableXRayTracing";
         private const string UseRuntimeErrorsKey = "UseRuntimeErrors";
+        private const string CollectSqlQueries = "CollectSqlQueries";
```
yogiraj07: Can we cover unit tests?

Issue #57, opened 2019-02-19
Entity doesn't exist in AsyncLocal exception when X-Ray tracing disabled


djluck:
Hi there,

When calling `GetAsyncResponseTraced`, I get the following exception:
```
Entity doesn't exist in AsyncLocal\n at Amazon.XRay.Recorder.Core.Internal.Context.AsyncLocalContextContainer.GetEntity() at Amazon.XRay.Recorder.Handlers.System.Net.Utils.RequestUtil.ProcessRequest(Uri uri, String method, Action`1 addHeaderAction) at Amazon.XRay.Recorder.Handlers.System.Net.Utils.RequestUtil.ProcessRequest(WebRequest request) at Amazon.XRay.Recorder.Handlers.System.Net.HttpWebRequestTracingExtension.GetAsyncResponseTraced(WebRequest request)
```

It seems that `RequestUtil.ProcessRequest` is attempting to access `AWSXRayRecorder.Instance.TraceContext.GetEntity` even though I have disabled x-ray tracing via my configuration (debugging my application confirms x-ray tracing is disabled). This seems like a bug to me- I wouldn't expect these instrumented methods to do anything if tracing was disabled.

yogiraj07:
Hi @djluck ,
Thanks for reporting the issue. I believe this [code](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/System.Net/Utils/RequestUtil.cs#L69) is causing the issue. It should be part of above if block. I would fix it soon.

A PR is also welcomed.

djluck:
I believe you are right @yogiraj07, I can put together a simple PR as I believe the fix will be just moving the referenced line inside the block that checks if X-ray is enabled.

jrlang1:
I am having the same bug with TraceableSqlCommand. Should I open a new issue

yogiraj07:
@jrlang1 , please feel free to open a separate issue. I will verify and do necessary changes. A PR is also welcomed. Thanks for posting the issue.

mbp:
Would be nice to get this fixed (new version on nuget). Impossible for us to use the library when we cannot ignore runtime errors.

yogiraj07:
@mbp, we are planning for a release soon. I will prioritize this release. Please stay tuned. On the side note, can you help me explain your use case ? What difficulties are you facing?  

Thanks for patience,
Yogi

mbp:
[...]

PR #58, opened 2019-02-26
Fixing bug in RequestUtil.ProcessRequest

djluck:
*Issue:* https://github.com/aws/aws-xray-sdk-dotnet/issues/57

*Description of changes:*
Don't access the trace context in `RequestUtil.ProcessRequest` when tracing is disabled. See  for more information.

### DIFF
--- a/sdk/test/UnitTests/HttpWebRequestTracingExtensionTests.cs
+++ b/sdk/test/UnitTests/HttpWebRequestTracingExtensionTests.cs
@@ -15,6 +15,7 @@
 // </copyright>
 //-----------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Net;
 using System.Threading.Tasks;
@@ -22,6 +23,7 @@ using Amazon.XRay.Recorder.Core;
 using Amazon.XRay.Recorder.Core.Internal.Entities;
 using Amazon.XRay.Recorder.Core.Internal.Utils;
 using Amazon.XRay.Recorder.Handlers.System.Net;
+using Amazon.XRay.Recorder.UnitTests.Tools;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Amazon.XRay.Recorder.UnitTests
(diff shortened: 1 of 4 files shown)


yogiraj07:
Hi @djluck ,
Can we add unit test with tracing disabled on the recorder for sync and async call for [HttpClient](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/test/UnitTests/HttpClientXRayTracingHandlerTests.cs) and [HttpWebRequest](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/test/UnitTests/HttpWebRequestTracingExtensionTests.cs).

djluck:
@yogiraj07 done. I also fixed an issue that would see test HTTP requests time out (response objects were not being disposed, so service point connections were not freed).

sdk/test/UnitTests/HttpWebRequestTracingExtensionTests.cs
```diff
@@ -60,19 +62,45 @@ public void TestGetResponseTraced()
...
+            AWSXRayRecorder.InitializeInstance(recorder: _recorder);
+# endif
+            Assert.IsTrue(AWSXRayRecorder.Instance.IsTracingDisabled());
 
-            var responseInfo = segment.Subsegments[0].Http["response"] as Dictionary<string, object>;
-            Assert.AreEqual(200, responseInfo["status"]);
-            Assert.IsNotNull(responseInfo["content_length"]);
+            var request = (HttpWebRequest)WebRequest.Create(URL);
+            
+            using (var response = request.GetResponseTraced() as HttpWebResponse)
```
yogiraj07: can we add unit test for `GetAsyncResponseTraced `

djluck: Pushed a fix for this.

yogiraj07:
Hi @djluck , 
Thank you for the PR.

djluck:
@yogiraj07 thanks for merging. When can I expect the latest release to be pushed out to nuget?

yogiraj07:
@djluck , is the fix blocking you? We usually maintain small release cycle.

djluck:
It is unfortunately (local dev does not work with tracing enabled OR disabled). There seems to be no workaround that I can think of.

yogiraj07:
@djluck ,

Can you please let me know your usecase ? What problem you are facing for local development? Will this PR fix the issue or not?

djluck:
I am working on an ASP.NET core application. It has two distinct environments: localdev (i.e. my local machine which does not have an X-Ray trace collection daemon) and lambda (has X-Ray). When running locally, either by directly running the application or running the integration tests, all traced HTTP calls fail with the exception outlined in the issue I filed. 
My solution will be to disable x-ray tracing locally and then enable it when running the application in AWS Lambda.

djluck:
I have been able to work around the issue for now. Thanks for your help @yogiraj07, I can wait for your normal release cycle.

Issue #59, opened 2019-03-12
Calling AddMetadata twice with the same key should overwrites previously recorded values


fodonnel:
According to the documentation (https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-dotnet-segment.html#xray-sdk-dotnet-segment-metadata), Calling AddMetadata twice with the same key should overwrites previously recorded values.

Currently the SDK throws an exception if you try to override the Metadata

PR #60, opened 2019-03-12
Bug fix - Calling AddMetadata twice with the same key should overwrit…

fodonnel:
…es previously recorded values

*Issue #, if available:*
#59 

*Description of changes:*
Calling AddMetadata twice with the same key should overwrites previously recorded values

### DIFF
--- a/sdk/src/Core/Internal/Entities/Entity.cs
+++ b/sdk/src/Core/Internal/Entities/Entity.cs
@@ -505,7 +505,7 @@ namespace Amazon.XRay.Recorder.Core.Internal.Entities
         /// <param name="value">The value.</param>
         public void AddMetadata(string nameSpace, string key, object value)
         {
-            _lazyMetadata.Value.GetOrAdd(nameSpace, new ConcurrentDictionary<string, object>()).Add(key, value);
+            _lazyMetadata.Value.GetOrAdd(nameSpace, new ConcurrentDictionary<string, object>())[key] = value;
         }
 
         /// <summary>
--- a/sdk/test/UnitTests/AWSXRayRecorderTests.cs
+++ b/sdk/test/UnitTests/AWSXRayRecorderTests.cs
@@ -184,7 +184,7 @@ namespace Amazon.XRay.Recorder.UnitTests
         }
 
         [TestMethod]
-        public void TestCreateThunsandSubsegmentsParallel()
+        public void TestCreateThousandSubsegmentsParallel()
         {
             _recorder.BeginSegment("parent", TraceId);
             Segment parent = (Segment)AWSXRayRecorder.Instance.TraceContext.GetEntity();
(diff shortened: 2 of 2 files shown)


sdk/src/Core/Internal/Entities/Entity.cs
```diff
@@ -505,7 +505,7 @@ public void AddMetadata(string key, object value)
         /// <param name="value">The value.</param>
         public void AddMetadata(string nameSpace, string key, object value)
         {
-            _lazyMetadata.Value.GetOrAdd(nameSpace, new ConcurrentDictionary<string, object>()).Add(key, value);
+            _lazyMetadata.Value.GetOrAdd(nameSpace, new ConcurrentDictionary<string, object>())[key] = value;
```
yogiraj07: Should we use [AddOrUpdate](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.addorupdate?view=netframework-4.5)

fodonnel: I tried that method first, based on its name its seemed like the right way to go!

Its on ConcurrentDictionary, not IDictionary. If I change the declaration to:
```
private readonly Lazy<ConcurrentDictionary<string, ConcurrentDictionary<string, object>>> _lazyMetadata = new Lazy<ConcurrentDictionary<string, ConcurrentDictionary<string, object>>>();
```

I am then forced me to change the signature of the Metadata Property to
```
public IDictionary<string, ConcurrentDictionary<string, object>> Metadata
{
    get
    {
        return _lazyMetadata.Value;
    }
}
````
which doesn't seem right.  Instead I can cast inside the AddMetadata method
```
var metadata = (ConcurrentDictionary<string, object>) _lazyMetadata.Value.GetOrAdd(nameSpace, new ConcurrentDictionary<string, object>());
metadata.AddOrUpdate(key, value, (k, v) => value);
```
which is better, still not great. And it bothers me that the I still couldn't do a simple `.AddOrUpdate(key, value);`, the best I can do is 
`.AddOrUpdate(key, value, (k, v) => value);`

Which caused me to rethink. Its odd signature and the fact that it is not available IDictionary leads me to believe its for a unique use case, situations where you want read a value from the cache and alter it (probably increment it) and save it in a thread safe operation.

That said I am happy to go with casting inside the method, or any better alternative if anybody knows one.

Comment (yogiraj07) [reply sdk/src/Core/Internal/Entities/Entity.cs]: Hi @fodonnel ,
Thank you for the explanation. I think lets not add casting logic and complicate it. However, can we add concurrency unit test on this method? The above link has a sample and the factor for parallel should be small used for testing purpose.

Comment (fodonnel) [reply sdk/src/Core/Internal/Entities/Entity.cs]: Hi @yogiraj07,

I am not sure there is a good way to do that. I tried stuff like
```
[TestMethod]
public void TestUpdateMetadataIsThreadSafe()
{
    _recorder.BeginSegment("metadata", TraceId);

    Parallel.For(0, 1000, i =>
    {
        _recorder.AddMetadata("key1", i);
    });

    var segment = AWSXRayRecorder.Instance.TraceContext.GetEntity();
    _recorder.EndSegment();

    Assert.AreEqual(1000, segment.Metadata["default"]["key1"]);
}
```
But that just fails, the metadata ends up containing a random number between 0 and 999, which makes sense, the last thread in wins and in this case that could be any thread.

The reason the test works in the linked article works is because they are using AddOrUpdate to increment (not replace) the value in the cache. So in their case, they called the method 1000 times - regardless of order the answer should be 1000.

I did find a explicit mention of using the indexer in the article it that helps (in the remarks section) https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2?view=netframework-4.5#remarks

| To do this  | Use this method |
| ------------- | ------------- |
| Store a key/value pair in the dictionary unconditionally, and overwrite the value of a key that already exists  | The indexer's setter: dictionary[key] = newValue  |

Issue #61, opened 2019-03-13
Add option to register XRay manifest from embedded resource


sergedomk:
Version: 2.5.0 (issue still exists on master)

I'm attempting to create a NuGet package for internal use that takes care of some of the XRay setup we are adding to a number of applications/services.

One of the requirements is adding some Secrets Manager calls to the whitelist. In fact, it was the primary reason for creating an internal library. That, and a DB profiler that worked on our PostgreSQL DB connections. Unfortunately, contentFiles are not pulled into projects that don't use project.json (https://blog.nuget.org/20160126/nuget-contentFiles-demystified.html - See section : Supported Project Types). In our case, this is almost all of them.

As, the SDK offers no option to pull in a manifest other than from file, I'm going to have to resort to taking my embedded manifest document from my NuGet package and streaming it to disk so I can then feed the path to the RegisterXRayManifest() method.

Ideally I would be able to register the embedded resource directly similar to how the DefaultAWSWhitelist.json resource is handled in InitWithDefaultAWSWhitelist() method in Handlers/AwsSdk/Internal/XRayPipelineHandler.cs or via a more generic approach that allows for passing a stream to RegisterXRayManifest().

yogiraj07:
Hi @sergedomk ,
Thank you for explaining the use case. 
I feel going with generic approach and adding a new method on `RegisterXRayManifest(string manifestName)` in this [class](https://github.com/aws/aws-xray-sdk-dotnet/blob/8053d17f191d40cbabb18b4cb20b6d6ff6cfb44f/sdk/src/Handlers/AwsSdk/AWSSDKHandler.cs#L63) will help.

I think we need to use [GetManifestResourceStream](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getmanifestresourcestream?redirectedfrom=MSDN&view=netframework-4.7.2#System_Reflection_Assembly_GetManifestResourceStream_System_String_) to read the stream. 

Please let me know your thoughts, I will take this as a feature request and a PR is always welcomed.

Thanks,
Yogi

sergedomk:
@yogiraj07 ,

Just to make sure I understand, I'm going to attempt to illustrate the approach I'm going to take. If it doesn't match with your thinking, I'd appreciate your feedback.

[...]

PR #63, opened 2019-03-16
Issue #61: Add option to register XRay manifest from embedded resource

sergedomk:
*Issue #61 :* Add option to register XRay manifest from embedded resource

*Description of changes:*

Added alternate `RegisterXRayManifest()` method which takes a `Stream` instead of a file path. To prevent having to leave the `Stream` open, modified `XRayPipelineCustomizer` to carry the `XRayPipelineHandler` rather than `Path` attribute so the manifest can be initialized on the call the `RegisterXRayManifest()` rather than the call to `Customize()`.

### DIFF
--- a/sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs
+++ b/sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs
@@ -89,6 +89,23 @@ namespace Amazon.XRay.Recorder.Handlers.AwsSdk.Internal
             }
         }
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="XRayPipelineHandler" /> class.
+        /// </summary>
+        /// <param name="stream"> stream for manifest which contains the operation parameter whitelist configuration.</param>
+        /// <exception cref="System.ArgumentNullException">Thrown when recorder is null.</exception>
+        public XRayPipelineHandler(Stream stream)
+        {
+            _recorder = AWSXRayRecorder.Instance;
+
+            if (_recorder == null)
+            {
+                throw new ArgumentNullException("recorder");
+            }
+
+            Init(_recorder, stream);
+        }
+
         private static bool TryReadPropertyValue(object obj, string propertyName, out object value)
         {
             value = 0;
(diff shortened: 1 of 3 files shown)


sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs
```diff
@@ -87,6 +87,23 @@ public XRayPipelineHandler(string path)
...
+        /// <exception cref="System.ArgumentNullException">Thrown when recorder is null.</exception>
+        public XRayPipelineHandler(Stream stream)
+        {
+            _recorder = AWSXRayRecorder.Instance;
+
+            if (_recorder == null)
+            {
+                throw new ArgumentNullException("recorder");
+            }
+
```
yogiraj07: Can we add a null check for the stream:
```
       if (stream == null)
            {
                _logger.DebugFormat("The provided stream is null, initializing with default AWS whitelist.");
                InitWithDefaultAWSWhitelist(_recorder);
            }
```

sdk/test/UnitTests/AWSSDKHandlerTests.cs
```diff
@@ -131,6 +131,16 @@ public void TestLoadServiceHandlerManifestWithDefaultConfigurationForAWSSDKHandl
             Assert.IsNotNull(handler.AWSServiceHandlerManifest);
         }
 
+        [TestMethod]
+        public void TestLoadServiceHandlerManifestWithDefaultConfigurationForAWSSDKHandlerAsStream()
+        {
+            using (Stream stream = new FileStream(_path, FileMode.Open, FileAccess.Read))
+            {
+                var handler = new XRayPipelineHandler(stream);
```
yogiraj07: A unit test for null stream would be nice.

sergedomk: Added!

sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs
sergedomk: Added!

PR #65, opened 2019-04-04
Honoring Context missing strategy for all TraceContext.GetEntity() call sites

yogiraj07:
*Description of changes:*
`AWSXRayRecorder.Instance.TraceContext.GetEntity()` call sites should be surrounded by `EntityNotAvailableException` catch block in order to honor ContextMissingStrategy configured to the X-Ray Recorder. Currently there are some calls that don't follow the pattern and this PR fies the issue.


sdk/test/UnitTests/HttpWebRequestTracingExtensionTests.cs
```diff
@@ -217,5 +217,32 @@ public async Task TestExceptionGetAsyncResponseTraced()
...
+#else
+            AWSXRayRecorder.InitializeInstance(recorder: _recorder);
+# endif
+            Assert.IsFalse(AWSXRayRecorder.Instance.IsTracingDisabled());
+
+            AWSXRayRecorder.Instance.ContextMissingStrategy = Core.Strategies.ContextMissingStrategy.LOG_ERROR;
+
+            _recorder.EndSegment();
+
+            // The test should not break. No segment is available in the context, however, since the context missing strategy is log error,
```
chrisradek: Is there a way to detect if the logger prints the correct message?

yogiraj07: Thats interesting point. However, there is no way to get the value from the logger message at the moment. I think, in future, we need to have custom logger implementation which has access to logger instance at AWSXRayRecorder level, to get the value of the message.

chrisradek: Any chance we can rerun now that we have appveyor? :)

yogiraj07: Sure, done.

PR #68, opened 2019-04-11
Initializes the static instance in AWSXrayRecorder to fix #67

### LINKED ISSUE #67: AwsXrayRecoroder trying to modify a non-concurrent collection without exclusive access?
Hi there I can't really provide code snippets for the project, but the context is that we are simply trying to call AWSXrayRecorder.Instance.BeginSubsegment("some name"); inside of a web api and it throws the following exception:

```
Error Message:
 System.InvalidOperationException : Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
Stack Trace:
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at Amazon.XRay.Recorder.Core.Internal.Emitters.JsonSegmentMarshaller..ctor()
   at Amazon.XRay.Recorder.Core.AWSXRayRecorder.get_Instance()
   at <redacted>
   at <redacted>
Error Message:
 System.InvalidOperationException : Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
Stack Trace:
   at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
   at System.Collections.Generic.Dictionary`2.set_Item(TKey key, TValue value)
   at Amazon.XRay.Recorder.Core.Internal.Emitters.JsonSegmentMarshaller..ctor()
   at Amazon.XRay.Recorder.Core.AWSXRayRecorder.get_Instance()
   at <redacted>
   at <redacted>
   at <redacted>
```

This is happening from inside a dependency injected class, so it would follow the following pattern (where everything is scoped as far as DI goes)
Controller --> Provider --> ProviderWithXraySubsegment

We do see the API tracing the segments started by making the Http Request just fine, but when we add subsegments, it seems to fail with this kind of an error. 

This will impact whether or not we use X-Ray at our business. We have also tried using the TraceMethod functions instead of creating subsegments.

jaybyrrd:
*Issue #, if available:* #67 

*Description of changes:* This adds a lock on the custom_exporters_table Dictionary and checks to see if it is null before moving forward. 

### DIFF
--- a/sdk/src/Core/AWSXRayRecorder.netcore.cs
+++ b/sdk/src/Core/AWSXRayRecorder.netcore.cs
@@ -32,7 +32,7 @@ namespace Amazon.XRay.Recorder.Core
     public class AWSXRayRecorder : AWSXRayRecorderImpl
     {
         private static readonly Logger _logger = Logger.GetLogger(typeof(AWSXRayRecorder));
-        static AWSXRayRecorder _instance;
+        static AWSXRayRecorder _instance = new AWSXRayRecorderBuilder().Build();
         public const String LambdaTaskRootKey = "LAMBDA_TASK_ROOT";
         public const String LambdaTraceHeaderKey = "_X_AMZN_TRACE_ID";
 


jaybyrrd:
Worth noting that I considered adding the following as well inside of the lock, as I had managed to show a stack trace where the custom_exporters_table was null:

```csharp
if(custom_exporters_table == null) {
    custom_exporters_table = new Dictionary<Type, ExporterFunc>();
}
```

But I didn't think it should be necessary and that the null reference error I was getting was related to the threading issue corrupting the object.

jaybyrrd:
Realizing this ran a ton of formatting on the file, going to force push a version that doesn't have all the formatting changes.

jaybyrrd:
Fixed formatting

yogiraj07:
Hi @jaybyrrd ,
I am doing some analysis of the exception. [RegisterExporter](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/ThirdParty/LitJson/JsonMapper.cs#L907) is called by [JsonSegmentMarshaller()](https://github.com/aws/aws-xray-sdk-dotnet/blob/bdb7471e2f14fc58da591321f08b3979729f0eb2/sdk/src/Core/Internal/Emitters/JsonSegmentMarshaller.cs#L39) and in turn called by [UdpSegmentEmitter()](https://github.com/aws/aws-xray-sdk-dotnet/blob/4cb1d10bd2de90a40f555d29f00cc6bcbb446ec7/sdk/src/Core/Internal/Emitters/UdpSegmentEmitter.cs#L45). This is called, technically only once during [AWSXRayRecorder](https://github.com/aws/aws-xray-sdk-dotnet/blob/bdb7471e2f14fc58da591321f08b3979729f0eb2/sdk/src/Core/AWSXRayRecorder.cs#L43) instance initialization which is a singleton object. 

Are you using [InitializeInstance()](https://github.com/aws/aws-xray-sdk-dotnet/tree/master#configuration)? I wonder if by any chance the AWSXRayRecorder instance is initialized twice at the same time. Can you do some trial on your side of adding lock instead to [InitializeInstance](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.netcore.cs#L81) or [Instance](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.netcore.cs#L150). For trial you can remove the lock in this PR and try the 2 combinations.

I see this error for the first time and could not reproduce locally. This PR fixes 2 types exception but need some more understanding of the issue.

jaybyrrd:
I am not calling Initialize Instance anywhere explicitly, I haven't had a chance but I can try locking the instance, but I don't think this should be the end game Dev experience.

yogiraj07:
@jaybyrrd , I meant adding lock inside [Instance](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.netcore.cs#L150) block i.e. at code level inside library itself. My suspect is by any chance [this](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.netcore.cs#L154) is causing issue, leading to multiple initialization of AWSXRayRecorder [instance](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.netcore.cs#L156) in non-thread safe way.

jaybyrrd:
@yogiraj07 Sorry for the delay, great. Will test. I just re-prioritized my time at work to work this out this week. Will get back to you soon.

jaybyrrd:
@yogiraj07 this seems to work and puts the lock in a spot that makes sure we enter that if block for the initialize from the getter only once.  Let me know if you want me to change the bounds of either lock.

sdk/src/Core/AWSXRayRecorder.netcore.cs
```diff
@@ -83,19 +86,22 @@ public static void InitializeInstance(IConfiguration configuration = null, AWSXR
...
+                recorder.XRayOptions = xRayOptions;
+                recorder = recorderBuilder.Build(recorder);
+            }
+            else
+            {
+                _logger.DebugFormat("Using default X-Ray recorder.");
+                recorder = recorderBuilder.Build(xRayOptions);
+            }
+
+            lock (_instance_lock)
```
yogiraj07: I don't think we need a lock here `InitializeInstance()` is called only once during configuring the web app.

sdk/src/Core/AWSXRayRecorder.netcore.cs
```diff
@@ -151,9 +157,12 @@ public static AWSXRayRecorder Instance
         {
             get
             {
-                if (_instance == null)
+                lock (_instance_lock)
```
yogiraj07: I would prefer avoiding lock here since this is in hot code path. Almost all X-Ray related operations call `Instance`. Here instead of lock, we can do an optimization. 
`_instance` is static instance. So we can initialize `_instance = new AWSXRayRecorderBuilder().Build();` [here](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.netcore.cs#L35) and eliminate ` if (_instance == null)` on the `get` of `Instance`.

Something similar to currently we do in `net45` platform. [Reference1](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.cs#L69), [Reference2](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorderImpl.cs#L40)

If you can modify this PR accordingly and do a testing on your environment, this will be helpful.
Avoiding locks will help in maintaining the current performance.

jaybyrrd: Perfect will post an update today.

jaybyrrd: Updated the pull request, this seems to solve it as well.

jaybyrrd: Looks like unit tests fail, taking a look today.

jaybyrrd: That's resolved now.

yogiraj07:
Hi @jaybyrrd ,
As a sanity check, the PR resolves the issue in your environment?

Issue #64, opened 2019-03-22
NullReferenceException in PopulateRequestAttributes


mbp:
In ASP.NET Core (2.2, but earlier versions also apply), you can get a NullReferenceException for requests that do not contain the host header:

```
Object reference not set to an instance of an object.

 at Microsoft.AspNetCore.Http.Extensions.UriHelper.GetDisplayUrl(HttpRequest request)
   at Amazon.XRay.Recorder.Handlers.AspNetCore.Internal.AWSXRayMiddleware.PopulateRequestAttributes(HttpRequest request, Dictionary`2 requestAttributes)
   at Amazon.XRay.Recorder.Handlers.AspNetCore.Internal.AWSXRayMiddleware.ProcessHTTPRequest(HttpContext context)
   at Amazon.XRay.Recorder.Handlers.AspNetCore.Internal.AWSXRayMiddleware.Invoke(HttpContext context)
```

Related issue: https://github.com/aspnet/AspNetCore/issues/2718

This can be solved by implementing your own GetDisplayUrl variant, which does the null check. You can basically just copy the one which will be used in .NET Core 3.0: https://github.com/aspnet/AspNetCore/commit/187e89f6f0ee67728a50f987d63234c1900130f

yogiraj07:
Hi @mbp ,
From the error stack, it looks like the issue is [here](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/AspNetCore/Internal/AWSXRayMiddleware.cs#L278). Thank you for the detail analysis.

It seems, for .NET Core, we need to implement our own logic (null safe) to get URL from the incoming request.

I will look into this soon. A PR is also welcomed and would be happy to review it. 

Best,
Yogi

yogiraj07:
Hi @mbp ,
Can you help me in reproducing this issue?
I tried spinning up ASP.NET Core 2.2 app locally, instrumented with X-Ray and curled the endpoint with following command:
`curl -v --http1.0 -H "Host:" <address>`

However, the server doesn't accept request with null host value and returns following response:

```
DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN""http://www.w3.org/TR/html4/strict.dtd">
<HTML><HEAD><TITLE>Bad Request</TITLE>
<META HTTP-EQUIV="Content-Type" Content="text/html; charset=us-ascii"></HEAD>
[...]

PR #72, opened 2019-04-16
Custom logic to get URL from incoming ASP.NET Core request

yogiraj07:
*Issue #64 , if available:*

*Description of changes:*
As per recommendation from ASP.NET Core team on this [issue](https://github.com/aspnet/AspNetCore/issues/2718), adding custom logic to fetch URL from the incoming ASP.NET Core request. 

###Testing
Tested locally for ASP.NET Core 2.2 web app.
For no host present ` curl -v --http1.0 -H "Host:" http://localhost:5000/`, `http` section in the trace generated by the logic :

```json
"http": {
              "request": {
                        "url": "http:///",
                        "method": "GET",
                        "user_agent": "curl/7.55.0",
                        "client_ip": "::1"
                },
                "response": {
                        "status": 200
                }
            },
```


sdk/src/Handlers/AspNetCore/Internal/AWSXRayMiddleware.cs
```diff
@@ -273,9 +275,32 @@ private static void PopulateRequestAttributes(HttpRequest request, Dictionary<st
             }
         }
 
-        private static string GetUrl(HttpRequest request)
+        // Implementing custom logic : https://github.com/aws/aws-xray-sdk-dotnet/issues/64
+        public static string GetUrl(HttpRequest request)
         {
-            return Microsoft.AspNetCore.Http.Extensions.UriHelper.GetDisplayUrl(request);
```
luluzhao: From my understanding, in stead of calling `GetDisplayUrl`, you implementing a way to construct the url by itself? Why you don't want to use the method when the request is not NULL.

yogiraj07: This is related to this [issue](https://github.com/aws/aws-xray-sdk-dotnet/issues/64) 

The NET Core 2.x throws null pointer exception on `GetDisplayUrl()` call, when `Host` header is not present in the incoming request. This is a bug from the downstream dependency which is fixed in NET Core 3.x (not yet GAed).

I talked to ASP NET Core community and they provided to copy paste the code : [link](https://github.com/aspnet/AspNetCore/issues/2718), see my comments in the last section. 

The request is not null is just a safety check, however the issue lies deeper inside the `request` object. Looking at the current situation, writing custom code gives fine grain control on null safety checks on individual parameters of the request instance.

luluzhao: I see. Once they fix the bug, are we going to keep our own implemented logic or using the new one they implemented?

yogiraj07: Once the ASP.NET Core back ports the change from NET Core 3.x to 2.x, I will upgrade the ASP.NET Core dependency and remove the custom logic. However the ETA is late June.

luluzhao: Cool, make sure we add this into our backlog to upgrade the new dependency once they fix it.

Issue #74, opened 2019-04-22
Error when pulling secret using custom Manifest


sergedomk:
Just pulled in 2.6.0 and got error attempting to pull secret from Secrets Manager.

I haven't researched this extensively yet, but it appears to be directly related to custom Manifest changes added in https://github.com/aws/aws-xray-sdk-dotnet/pull/63.

Given that I was responsible for this change, I feel I must apologize for not catching it.

Received the following error regardless of whether the custom manifest was added as a path or stream.

```
AmazonXRayClient 20|2019-04-22T09:10:01.942Z|ERROR|An exception of type HttpErrorResponseException was handled in ErrorHandler. --> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
   at Amazon.Runtime.HttpWebRequestMessage.GetResponseAsync(CancellationToken cancellationToken) in D:\JenkinsWorkspaces\trebuchet-stage-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\HttpHandler\_mobile\HttpRequestMessageFactory.cs:line 539
   at Amazon.Runtime.Internal.HttpHandler`1.InvokeAsync[T](IExecutionContext executionContext) in D:\JenkinsWorkspaces\trebuchet-stage-release\AWSDotNetPublic\sdk\src\Core\Amazon.Runtime\Pipeline\HttpHandler\HttpHandler.cs:line 175
   at Amazon.Runtime.Internal.Unmarshaller.InvokeAsync[T](IExecutionContext executionContext)
   at Amazon.Runtime.Internal.ErrorHandler.InvokeAsync[T](IExecutionContext executionContext)
AmazonXRayClient 21|2019-04-22T09:10:02.041Z|ERROR|AmazonSecretsManagerException making request GetSecretValueRequest to https://secretsmanager.us-east-1.amazonaws.com/. Attempt 1. --> Amazon.SecretsManager.AmazonSecretsManagerException: Missing Authentication Token ---> Amazon.Runtime.Internal.HttpErrorResponseException: Exception of type 'Amazon.Runtime.Internal.HttpErrorResponseException' was thrown.
[...]

PR #75, opened 2019-04-23
Fixing AWS SDK handler for registering all clients using custom AWS Service manifest file

yogiraj07:
*Issue #74 *

*Description of changes:*
PR #61 changed the way `XRayPipelineHandler` instance [created](https://github.com/aws/aws-xray-sdk-dotnet/pull/63/files#diff-9f4840a8c388c0e3929a12615bdf65e2R675), which caused regression. This PR maintains the original behavior as required by the interface.  


sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs
```diff
@@ -104,17 +94,65 @@ public XRayPipelineHandler(Stream stream)
...
+        /// <param name="awsServiceManifest">Instance of <see cref="AWSServiceHandlerManifest"/></param>
+        public XRayPipelineHandler(AWSServiceHandlerManifest awsServiceManifest)
+        {
+            _recorder = AWSXRayRecorder.Instance;
+
+            if (_recorder == null)
             {
-                _logger.DebugFormat("The provided stream is null, initializing with default AWS whitelist.");
-                InitWithDefaultAWSWhitelist(_recorder);
+                throw new ArgumentNullException("recorder");
```
chrisradek: It looks like if the recorder is missing when the `XRayPipelineHandler` is instantiated we'll throw an exception that isn't caught. Is this the same behavior as before the custom manifest was supported, or should it be logged instead?

yogiraj07: This is a new constructor and inline with existing [constructor](https://github.com/aws/aws-xray-sdk-dotnet/blob/c9189893d31d692d4f25d6adb4b0f2143d41dd5e/sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs#L70) which maintains the same behavior.

PR #83, opened 2019-06-05
Added Subsegment streaming strategy.

srprash:
*Issue #, if available:*

*Description of changes:*
Added an interface `IStreamingStrategy` to enable customization of streaming strategies. Defaults to the DefaultStreamingStrategy if custom strategy is not provided.

### DIFF
--- a/sdk/src/Core/Strategies/DefaultStreamingStrategy.cs
+++ b/sdk/src/Core/Strategies/DefaultStreamingStrategy.cs
@@ -1,4 +1,5 @@
-﻿using Amazon.XRay.Recorder.Core.Internal.Entities;
+﻿using System;
+using Amazon.XRay.Recorder.Core.Internal.Entities;
 using Amazon.XRay.Recorder.Core.Internal.Emitters;
 using Amazon.XRay.Recorder.Core.Sampling;
 
@@ -30,6 +31,10 @@ namespace Amazon.XRay.Recorder.Core.Strategies
         /// <param name="maxSubsegmentSize"></param>
         public DefaultStreamingStrategy(long maxSubsegmentSize)
         {
+            if(maxSubsegmentSize < 0)
+            {
+                throw new ArgumentException("maxSubsegmentSize cannot be a negative number.");
+            }
             MaxSubsegmentSize = maxSubsegmentSize;
         }
 
@@ -38,7 +43,7 @@ namespace Amazon.XRay.Recorder.Core.Strategies
         /// </summary>
         /// <param name="entity">Instance of <see cref="Entity"/></param>
         /// <returns>True if the subsegments are streamable.</returns>
-        public bool ShouldStreamSubsegments(Entity entity)
+        public bool ShouldStream(Entity entity)
         {
             return entity.Sampled == SampleDecision.Sampled && entity.RootSegment != null && entity.RootSegment.Size >= MaxSubsegmentSize;
         }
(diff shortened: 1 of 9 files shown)


sdk/src/Core/Strategies/DefaultStreamingStrategy.cs
```diff
@@ -0,0 +1,77 @@
...
+        public DefaultStreamingStrategy() : this(DefaultMaxSubsegmentSize)
+        {
+            
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DefaultStreamingStrategy"/> class.
+        /// </summary>
+        /// <param name="maxSubsegmentSize"></param>
+        public DefaultStreamingStrategy(long maxSubsegmentSize)
```
yogiraj07: We should do non-negative number check while setting this value. If negative number throw relevant exception. Also include unit test for the same.

sdk/src/Core/Strategies/IStreamingStrategy.cs
```diff
@@ -0,0 +1,42 @@
...
+    /// Interface of streaming strategy which is used to determine when and how the subsegments will be streamed.
+    /// </summary>
+    public interface IStreamingStrategy
+    {
+        /// <summary>
+        /// Determines whenther or not the provided segment requires any subsegment streaming.
+        /// </summary>
+        /// <param name="input">An instance of <see cref="Segment"/>.</param>
+        /// <returns>true if the segment should be streamed.</returns>
+        bool ShouldStreamSubsegments(Entity entity);
```
yogiraj07: I think we should rename this to `ShouldStream`, because we check for `Segment` whether its `Subsegment/s` should be streamed.

sdk/src/Core/Strategies/IStreamingStrategy.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        /// <summary>
+        /// Streams subsegments of instance of <see cref="Entity"/>.
+        /// </summary>
+        /// <param name="entity">Instance of <see cref="Entity"/>.</param>
+        /// <param name="emitter">Instance if <see cref="ISegmentEmitter"/>.</param>
+        void StreamSubsegments(Entity entity, ISegmentEmitter emitter);
```
yogiraj07: Lets keep the method name generic to `Stream()`.

sdk/test/UnitTests/AWSXRayRecorderTests.cs
```diff
@@ -1019,7 +1020,20 @@ public void TestInitializeInstanceWithRecorder3() // setting custom daemon addre
 #endif
             Assert.AreEqual(typeof(DummyTraceContext), AWSXRayRecorder.Instance.TraceContext.GetType()); // Custom trace context
             recorder.Dispose();
+            AWSXRayRecorder.Instance.Dispose();
+        }
+
+        [TestMethod]
+        public void TestDefaultStreamingStrategy()
+        {
+            IStreamingStrategy defaultStreamingStrategy = new DefaultStreamingStrategy(50);
```
yogiraj07: Lets split this unit test into separate unit tests:
1. DefaultStreamingStrategy() -> check default value is set 
2. DefaultStreamingStrategy(45) -> value other other than default value is correctly set
3. DefaultStreamingStrategy(-10) -> negative value is set

PR #87, opened 2019-06-16
Readme fix - Using correct namespace and nuget package

srprash:
*Issue #, if available:* 78

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,7 +442,7 @@ public string FunctionHandler(string input, ILambdaContext context)
 
 ### ASP.NET Core on AWS Lambda (.NET Core)
 
-We support instrumenting ASP.NET Core web app on Lambda. Please follow the steps of [ASP.NET Core](https://github.com/aws/aws-xray-sdk-dotnet/tree/release#aspnet-core-framework-net-core) instrumentation.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -97,6 +97,13 @@ AWSXRayRecorder.InitializeInstance(configuration); // pass IConfiguration object
 1. You should configure this before initialization of `AWSXRayRecorder` instance and using any AWS X-Ray methods.  
 2. If you manually need to configure `IConfiguration` object refer: [Link](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?tabs=basicconfiguration)  
 3. For more information on configuration, please refer : [Link](https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-dotnet-configuration.html)
+4. You can also set up the `AWSXRayRecorder` instance programmatically instead of a configuration file by simply using the `AWSXRayRecorderBuilder` class. For example, if you want to create a recorder with your custom `IStreamingStrategy`, you can do it through the following code.
```
yogiraj07: Instead in the `Note` section, we can create a separate topic under `Configuration` say `Configuring Recorder` or something.  Also this is configuring X-Ray recorder instance programmatically better then "set up". We should explicitly state `CustomStreamingStrategy` should implement `IStreamingStrategy`. Also add another configuration in the example for example : `new AWSXRayRecorderBuilder().WithStreamingStrategy(new CustomStreamingStrategy()).WithSegmentEmitter(new CustomEmitter())Build();`, so it becomes clear that multiple configuration is possible on the same instance.

PR #90, opened 2019-07-05
Whitelist SNS for recording the TopicArn on Publish

srprash:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/sdk/src/Handlers/AwsSdk/DefaultAWSWhitelist.json
+++ b/sdk/src/Handlers/AwsSdk/DefaultAWSWhitelist.json
@@ -81,7 +81,6 @@
 //   SimpleSystemsManagement
 //   SimpleWorkflow
 //   Snowball
-//	 SNS
 //   SQS
 //   StepFunctions
 //   StorageGateway
@@ -229,7 +228,7 @@
             }
         }
     },
-    "SNS": {
+    "SimpleNotificationService": {
 		"operations": {
 			"Publish": {
 				"request_parameters": [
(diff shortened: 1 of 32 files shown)


sdk/src/Handlers/AwsSdk/DefaultAWSWhitelist.json
```diff
@@ -228,7 +228,16 @@
             }
         }
     },
-    "SQS": {
+    "SimpleNotificationService": {
```
yogiraj07: sanity check : Should this be "SNS"?

srprash: The serviceName which we receive from the AWS SDK is "SimpleNotificationService" which we then match in our whitelist here: https://github.com/aws/aws-xray-sdk-dotnet/blob/d5ce31fad3517e0e1284e94a0a530f9cd3f17291/sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs#L552

Therefore, this needs to be "SimpleNotificationService" and not "SNS".

srprash: Sure. So for testing purposes, I created a SNS Topic. Using a small .Net application (instrumented to register X-Ray for all AWS services), and the AWS SDK for .Net, I made a `Publish` request to the topic with a message. I was able to see the generated subsegment on the X-Ray console in a trace. The JSON for the `SimpleNotificationService` subsegment now had a `topic_arn` field in addition to the `request_id`, `region`, and `operation` fields which we capture for all the services irrespective of their whitelist status. Also verified that using "SNS" as service name in the [DefaultAWSWhitelist.json](https://github.com/srprash/aws-xray-sdk-dotnet/blob/net45_add_sql_query_to_subsegment/sdk/src/Handlers/AwsSdk/DefaultAWSWhitelist.json) didn't work by putting breakpoints.

PR #92, opened 2019-07-11
Added SetUser to Segment

srprash:
*Issue #, if available:*

*Description of changes:*
Added a User property to Segment.
Added public methods to get and set the User.
Moved HasStreamed property to Entity since it is being used by both Segment and SubSegment now.
Modified JsonSegmentMarshaller to stream user for a segment.
Added unit tests. 

### DIFF
--- a/sdk/src/Core/Internal/Entities/Segment.cs
+++ b/sdk/src/Core/Internal/Entities/Segment.cs
@@ -21,6 +21,7 @@ using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
+using Amazon.XRay.Recorder.Core.Exceptions;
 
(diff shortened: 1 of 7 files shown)


sdk/src/Core/Internal/Entities/Segment.cs
```diff
@@ -162,5 +169,41 @@ public void SetEndTime(DateTime timestamp)
         {
             EndTime = TimeStamp.ToUnixSeconds(timestamp);
         }
+
+        /// <summary>
+        /// Checks if the segment has been streamed already
+        /// </summary>
+        /// <exception cref="AlreadyEmittedException">The segment has been already streamed and no further operation can be performed on it.</exception>
+        public void HasAlreadyStreamed()
```
yogiraj07: the modifier should be `private`?

Issue #76, opened 2019-05-06
Custom start and end date for Segment and Subsegment


okubiak:
Hi,

I'm currently facing a little issue trying to profile Elasticache Redis commands.
I'm using StackExchange.Redis client and the profiling system included is not really friendly with the AWSXRayRecorder design.

All what I can have is a list of commands elapsed time.
I tried to set the start and end times for a subsegment but `AWSXRayRecorder.Instance.EndSubsegment();` is overriding the end time with utcnow (which is fair).
For now I just added the elapsed time in the subsegment metadata but it will be complicate to investigate later in XRay console.
```
recorder.BeginSubsegment("Redis Commands");
recorder.SetNamespace("Database");
recorder.AddMetadata(nameof(command.Command), command.Command);
recorder.AddMetadata(nameof(command.ElapsedTime), command.ElapsedTime);
AWSXRayRecorder.Instance.EndSubsegment();
```

So my first question is: Do you think that customizing start and end time would be a solution at some point?
As StackExchange.Redis does not provide any callback where I could call a `BeginSubsegment` and `EndSubsegment`...

Second question: Do you have any other recommendation to profile Redis?

Regards

yogiraj07:
Hi @okubiak ,
1. Currently [BeginSegment()](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorderImpl.cs#L146) and [EndSegment()](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorderImpl.cs#L188) accepts custom start and end time for the `Segment`. For your use case, we need to add similar support for `BeginSubsegment()` and `EndSubsegment()` which will ease the user experience.

2. For second question, `StackExchange.Redis` is a custom client and I think surrounding the call with custom `Subsegment` is the way to go. 

I will keep this issue open and tag it as a feature request. You are welcome to submit a PR.

Thanks,
Yogi

srprash:
Hi @okubiak ,
This feature has been implemented in v2.7.0
Please refer to the [Changelog](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/CHANGELOG.md) for details.

Thanks,
Prashant

PR #93, opened 2019-07-15
Custom timestamp for subsegment

srprash:
*Issue #, if available:* [76](https://github.com/aws/aws-xray-sdk-dotnet/issues/76)

*Description of changes:*
Added timestamp parameter to BeginSubsegment and EndSubsegment.
Added Unit tests for the same.

### DIFF
--- a/sdk/src/Core/AWSXRayRecorder.netcore.cs
+++ b/sdk/src/Core/AWSXRayRecorder.netcore.cs
@@ -173,9 +173,10 @@ namespace Amazon.XRay.Recorder.Core
         /// Begin a tracing subsegment. A new segment will be created and added as a subsegment to previous segment/subsegment.
         /// </summary>
         /// <param name="name">Name of the operation</param>
+        /// <param name="timestamp">Sets the start time of the subsegment</param>
         /// <exception cref="ArgumentNullException">The argument has a null value.</exception>
         /// <exception cref="EntityNotAvailableException">Entity is not available in trace context.</exception>
-        public override void BeginSubsegment(string name)
+        public override void BeginSubsegment(string name, DateTime? timestamp = null)
         {
             try
             {
@@ -187,11 +188,11 @@ namespace Amazon.XRay.Recorder.Core
 
                 if (IsLambda())
                 {
-                    ProcessSubsegmentInLambdaContext(name);
+                    ProcessSubsegmentInLambdaContext(name, timestamp);
                 }
                 else
                 {
-                    AddSubsegment(new Subsegment(name));
+                    AddSubsegment(new Subsegment(name), timestamp);
                 }
             }
             catch (EntityNotAvailableException e)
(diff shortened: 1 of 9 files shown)


sdk/test/UnitTests/AWSXRayRecorderTests.cs
```diff
@@ -1052,6 +1052,35 @@ public void TestDefaultStreamingStrategyWithNegativeValue()
             Assert.ThrowsException<ArgumentException>(() => new AWSXRayRecorderBuilder().WithStreamingStrategy(new DefaultStreamingStrategy(-100)));
         }
 
+        [TestMethod]
+        public void TestBeginSubsegmentWithCustomTime()
```
yogiraj07: Sanity check : unit test that checks begin/end subsegment if no value is passed? The assertion should be on the current time.

sdk/src/Core/AWSXRayRecorder.netcore.cs
```diff
@@ -258,19 +259,27 @@ internal void AddFacadeSegment(String name = null)
...
+        private void AddSubsegmentInLambdaContext(string name, DateTime? timestamp = null)
         {
             // If the request is not sampled, the passed subsegment will still be available in TraceContext to
             // stores the information of the trace. The trace information will still propagated to 
             // downstream service, in case downstream may overwrite the sample decision.
             Entity parentEntity = TraceContext.GetEntity();
+            Subsegment subsegment = new Subsegment(name);
             parentEntity.AddSubsegment(subsegment);
             subsegment.Sampled = parentEntity.Sampled;
-            subsegment.SetStartTimeToNow();
```
yogiraj07: Make sure unit tests are covered for Lambda workflow.

sdk/src/Core/AWSXRayRecorder.netcore.cs
```diff
@@ -203,12 +204,12 @@ public override void BeginSubsegment(string name)
...
         /// Begin a tracing subsegment. A new subsegment will be created and added as a subsegment to previous facade segment or subsegment.
         /// </summary>
-        private void ProcessSubsegmentInLambdaContext(string name)
+        private void ProcessSubsegmentInLambdaContext(string name, DateTime? timestamp = null)
         {
             if (!TraceContext.IsEntityPresent()) // No facade segment available and first subsegment of a subsegment branch needs to be added
             {
                 AddFacadeSegment(name);
-                AddSubsegmentInLambdaContext(new Subsegment(name));
+                AddSubsegmentInLambdaContext(name, timestamp);
```
yogiraj07: Avoid duplicate code -  `AddSubsegmentInLambdaContext(new Subsegment(name,timestamp))` will do necessary setting of the start time. Same thing for `AddSubsegment()`. In the current function definitions you just remove code line that sets the start time.

sdk/src/Core/Sampling/TimeStamp.cs
```diff
@@ -94,7 +94,7 @@ internal bool IsAfter(TimeStamp timeStamp)
 
         private static decimal GetUnixSeconds(DateTime utcNow)
```
yogiraj07: Was this a bug?

srprash: Yes. It was always returning the current time as unix seconds.
It is a helper method primarily called by SetStartTime(timestamp) and SetEndTime(timestamp)

sdk/src/Core/AWSXRayRecorder.netcore.cs
srprash: Currently we don't set properties other than `name` for new `Subsegment` instance.

Issue #97, opened 2019-08-17
How about updating Least .NET Framework version  to .NET452


enif-lee:
Hello there.

I have tried to contribute for adding some extensions (ex: NpgsqlHandler, MysqlHandler, Entity Framework Support). But The solution failed to build because of .net framework version.

[.NET45 is officially end](https://devblogs.microsoft.com/dotnet/support-ending-for-the-net-framework-4-4-5-and-4-5-1/). So I can't download developer pack for 45. from [.net download](https://dotnet.microsoft.com/download/dotnet-framework).

However, To learn code base of this project, I raise .net version to 452, finally I've succeed to build whole projects.

I would like to register this as a PR, What do you think about raising minimum .net framework version to 452?

Thank you.
Best Regards,

srprash:
Hi @enif-lee ,
A PR is always welcomed. Feel free to create one. We will review it and get it merged.

Thanks.

PR #101, opened 2019-09-02
Update .net framework version to 4.5.2

enif-lee:
*Issue #97 *

*Change symbols of .NET Framework 4.5 to 4.5.2*


enif-lee:
This changes is something wrong. In my thinking, New Line Syntax is changes. 

Could you tell me which new line syntax is lf or crlf? and how about using [.editorconfig](https://editorconfig.org/)?

srprash:
Hi @enif-lee ,
Thanks for taking the time to submit this PR. I see that currently most of the files are with LF line endings but there are few CRLFs too. This is because the repo has been updated from different OS. I think using a `.gitattribute` file can be a better option and we would be able to define line endings at the repo level which is convenient and uniform for all the contributors.

enif-lee:
@srprash Yeah, The new line configuration of attribute is already done on #132 and it seems to be working fine.

srprash:
Great! We can get this merged and released for our next major version. :)

PR #123, opened 2020-03-20
Added support for tracing SQL query through Entity Framework Core

lupengamzn:
*Issue:*

Added support for tracing SQL query through Entity Framework Core 3.0 and above. (.NET Core)

*Description of changes:*

Entity Framework 3.0 starts to include `DbCommandInterceptor` for interception of database operations.

https://docs.microsoft.com/en-us/ef/core/what-is-new/ef-core-3.0/#interception-of-database-operations

https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.diagnostics.dbcommandinterceptor?view=efcore-3.0

Integrated `DbCommandInterceptor` with AWS XRay SDK to enable tracing SQL query through EF Core 3.0 and above.  

* Extract `database_type` from `CommandEventData.Context`.
* Subsegment name is correctly collected via `DbCommand` through EF Core.
* `CollectSqlQueries` option can be enabled locally or globally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,25 +395,24 @@ For how to start with Entity Framework Core in an ASP.NET Core web app, please t
 
 #### Setup
 
-In order to trace SQL query, you can register your `DbContext` with `EFInterceptor` in the `ConfigureServices` method in `startup.cs` file. 
+In order to trace SQL query, you can register your `DbContext` with `AddXRayInterceptor()` accordingly in the `ConfigureServices` method in `startup.cs` file. 
 
 For instance, when dealing with MySql server using Nuget: [Pomelo.EntityFrameworkCore.MySql](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql) (V 3.1.1). 
 
 ```csharp
-public void ConfigureServices(IServiceCollection services){ 
-
-    services.AddDbContext<your_DbContext>(options => options.UseMySql(your_connectionString).AddInterceptors(new EFInterceptor()));
-
+public void ConfigureServices(IServiceCollection services)
+{ 
+    services.AddDbContext<your_DbContext>(options => options.UseMySql(your_connectionString).AddXRayInterceptor());
 }
 ```
 
-Alternatively, you can register `EFInterceptor` in the `Onconfiguring` method in your `DbContext` class. Below we are using Nuget: [Microsoft.EntityFrameworkCore.Sqlite](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Sqlite) (V 3.1.2)
+Alternatively, you can register `AddXRayInterceptor()` in the `Onconfiguring` method in your `DbContext` class. Below we are using Nuget: [Microsoft.EntityFrameworkCore.Sqlite](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Sqlite) (V 3.1.2)
 
 ```csharp
 public class your_DbContext : DbContext 
 {
 	protected override void OnConfiguring(DbContextOptionsBuilder options)
-    => options.UseSqlite(your_connectionString).AddInterceptors(new EFInterceptor());
+    => options.UseSqlite(your_connectionString).AddXRayInterceptor();
 }
 ```
 
@@ -433,13 +432,12 @@ If you want to enable this feature, it can be done in two ways. First, by settin
 }
 ```
 
-Secondly, you can set the `collectSqlQueries` parameter in the `EFInterceptor` instance as **true** to collect the SQL query text. If you set this parameter as **false**, it will disable the `collectSqlQueries` feature for this `EFInterceptor` instance.
+Secondly, you can set the `collectSqlQueries` parameter in the `AddXRayInterceptor()` as **true** to collect the SQL query text. If you set this parameter as **false**, it will disable the `collectSqlQueries` feature for this `AddXRayInterceptor()`.
 
 ```csharp
-public void ConfigureServices(IServiceCollection services){
-
-    services.AddDbContext<your_DbContext>(options => options.UseMySql(your_connectionString).AddInterceptors(new EFInterceptor(true)));
-
+public void ConfigureServices(IServiceCollection services)
+{
+    services.AddDbContext<your_DbContext>(options => options.UseMySql(your_connectionString).AddXRayInterceptor(true));
 }
 ```
 
@@ -449,7 +447,7 @@ Or
 public class your_DbContext : DbContext 
 {
 	protected override void OnConfiguring(DbContextOptionsBuilder options)
-    => options.UseSqlite(your_connectionString).AddInterceptors(new EFInterceptor(true));
+    => options.UseSqlite(your_connectionString).AddXRayInterceptor(true);
 }
 ```
  
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ using (var command = new TraceableSqlCommand("SELECT * FROM products", connectio
 
 AWS XRay SDK for .NET Core provides interceptor for tracing SQL query through Entity Framework Core (>=3.0).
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -383,6 +383,76 @@ using (var command = new TraceableSqlCommand("SELECT * FROM products", connectio
...
+#### Setup
+
+In order to trace SQL query, you can register your `DbContext` with `EFInterceptor` in the `ConfigureServices` method in `startup.cs` file. 
+
+For instance, when dealing with MySql server using Nuget: [Pomelo.EntityFrameworkCore.MySql](https://www.nuget.org/packages/Pomelo.EntityFrameworkCore.MySql) (V 3.1.1). 
+
+```csharp
+public void ConfigureServices(IServiceCollection services){ 
+
+    services.AddDbContext<your_DbContext>(options => options.UseMySql(your_connectionString).AddInterceptors(new EFInterceptor()));
```
normj: To improve discover ability I would suggest adding an extension method to `DbContextOptionsBuilder` called `AddXRayTracing` or something which do the work of calling  `AddInterceptors` with the `EFInterceptor`. That way Intellisense would pretty tell the user how to enable the feature instead of relying on the developer to understand about interceptors and what class that should create and pass in. This approach is similar to the `UseXRay` extension method used in the ASP.NET Core handler.

sdk/src/Handlers/EntityFramework/AWSXRayRecorder.Handlers.EntityFramework.csproj
```diff
@@ -0,0 +1,35 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <Company>Amazon.com, Inc</Company>
+    <Product>Amazon Web Service X-Ray Recorder</Product>
+    <Copyright>Copyright 2017-2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.</Copyright>
```
normj: 2020?

sdk/src/Handlers/EntityFramework/AWSXRayRecorder.Handlers.EntityFramework.csproj
```diff
@@ -0,0 +1,35 @@
...
+    <AssemblyOriginatorKeyFile>../../../../buildtools/local-development.snk</AssemblyOriginatorKeyFile>
+    <DocumentationFile>bin\$(Configuration)\$(TargetFramework)\$(AssemblyName).xml</DocumentationFile>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>1701;1702;1591;</NoWarn>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore" Version="2.0.0" />
```
normj: I don't think you need the Microsoft.AspNetCore.Diagnostics.EntityFrameworkCore dependency.

sdk/src/Handlers/EntityFramework/EFInterceptor.cs
```diff
@@ -0,0 +1,389 @@
...
+            // Remove sensitive information from connection string
+            connectionStringBuilder.Remove("Password");
+
+            // Do a pre-check for UserID since in the case of TrustedConnection, a UserID may not be available.
+            var user_id = GetConnectionValue(connectionStringBuilder, "User ID");
+            if (user_id != null)
+            {
+                _recorder.AddSqlInformation("user", user_id.ToString());
+            }
+
```
matthewdouglas: Some connection strings will use a different key, e.g. `User` or `Username`.

lupengamzn: @matthewdouglas Exactly, they do have different format options

sdk/src/Handlers/EntityFramework/EFInterceptor.cs
```diff
@@ -22,23 +22,11 @@
...
 using System.Threading;
 using System.Threading.Tasks;
-using System.Text.RegularExpressions;
-using Microsoft.EntityFrameworkCore;
 
 namespace Amazon.XRay.Recorder.Handlers.EntityFramework
 {
     public class EFInterceptor : DbCommandInterceptor
     {
-        // Some database providers may not support Entity Framework Core 3.0 and above now
```
lupengamzn: Refracted to EFUtil.cs

sdk/src/Handlers/EntityFramework/EFInterceptor.cs
```diff
@@ -329,65 +317,13 @@ protected virtual void CollectSqlInformation(CommandEventData eventData)
             }
         }
 
-        private string GetDataBaseType(DbContext context)
```
lupengamzn: Refracted to EFUtil.cs

sdk/src/Handlers/EntityFramework/EFInterceptor.cs
```diff
@@ -329,65 +317,13 @@ protected virtual void CollectSqlInformation(CommandEventData eventData)
...
-            return value;
-        }
-
         /// <summary>
         /// Builds the name of the subsegment in the format database@datasource
         /// </summary>
         /// <param name="command">Instance of <see cref="DbCommand"/>.</param>
         /// <returns>Returns the formed subsegment name as a string.</returns>
         private string BuildSubsegmentName(DbCommand command)
-            => command.Connection.Database + "@" + RemovePortNumberFromDataSource(command.Connection.DataSource);
```
lupengamzn: Refracted to EFUtil.cs

sdk/src/Handlers/EntityFramework/AWSXRayInterceptorExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+//      or in the "license" file accompanying this file. This file is distributed
+//      on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+//      express or implied. See the License for the specific language governing
+//      permissions and limitations under the License.
+// </copyright>
+//-----------------------------------------------------------------------------
+
+using Microsoft.EntityFrameworkCore;
+
+namespace Amazon.XRay.Recorder.Handlers.EntityFramework
```
normj: For extensions like this we generally put them in the same namespace as the type we are putting the extension method on. In this case the namespace should be `Microsoft.EntityFrameworkCore`. That way the methods light up as soon as the add the NuGet package. This was done in X-Ray's extension methods for [IApplicationBuilder](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Handlers/AspNetCore/AWSXRayMiddlewareExtensions.cs).

lupengamzn: Will change it later

lupengamzn:
Switched over to [#124 ](https://github.com/aws/aws-xray-sdk-dotnet/pull/124)

sdk/src/Handlers/EntityFramework/EFUtil.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            string databaseProvider = context?.Database?.ProviderName;
+
+            // Need to check if the context and its following parameter is null or not to avoid exception
+            if (string.IsNullOrEmpty(databaseProvider))
+            {
+                return DefaultDatabaseType;
+            }
+
+            string value = null;
```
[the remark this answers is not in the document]
lupengamzn: If the `value` is not found, then this method will return the database provider name, which will not cause NPEs.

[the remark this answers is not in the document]
lupengamzn: > I would recommend initializing `value` to an empty string to avoid NPEs down the line.

If the value is not found, then this method will return the database provider name, which will not cause NPEs.

PR #124, opened 2020-03-24
Tracing Support for Entity Framework Core 3.0 and above

srprash:
*Issue #, if available:*

*Description of changes:*


sdk/test/UnitTests/EfCoreTests.cs
```diff
@@ -0,0 +1,120 @@
...
+            var segment = recorder.TraceContext.GetEntity();
+            Assert.AreEqual(3, segment.Subsegments.Count); //2 subsegments from validating the db and 1 from the actual query
+            var query_subsegment = segment.Subsegments[2];
+            AssertQueryCollected(query_subsegment);
+
+            recorder.EndSegment();
+        }
+
+        [TestMethod]
+        public void Test_EFCore_unsuccessful_query()
```
srprash: This is still a test for successful query capture from EFCore, only with collect_sql_queries as false. We should also test catching of exceptions when a DBCommand fails and invokes the `CommandFailed` method of the `EFInterceptor`. The test should validate that the exception is captured alongwith other information in the subsegment.

emonhaider:
@srprash I see this is merged to master, was the nuget published?

lupengamzn:
Hi @emonhaider, the Nuget will be published soon in this week

Issue #102, opened 2019-09-12
.Net SDK release v3.0.0


srprash:
This is a tracking issue for the changes going into the next major version release **`v3.0.0`**:
### 1.  Fixed a typo in Reservoir [PR #62]:
Changes the class and methods from `Reservior` to `Reservoir`.

### 2. Update .net framework version to 4.5.2 [PR #101]:
Current SDK targets `net45` as the `TargetFramework` for its projects. Since, the support for .Net Framework 4, 4.5 and 4.5.1 has [officially ended](https://devblogs.microsoft.com/dotnet/support-ending-for-the-net-framework-4-4-5-and-4-5-1/), we plan to target the `net452` version. This would break customers who are using the SDK with applications targeting `net45`.

jdavis-leasehawk:
How about .NET Core 3.1 (LTS) support?
Microsoft is soon to abandon .NET Core 3.0.

jon-armen:
Re: Item 4: Update IPlugin interface to use async methods.

I don't see how this will be possible.  I took a look into what it would take, and it leads down to AWSXRayRecorderBuilder().Build(), which is a core piece of initialization, and would require making much of the library async, including places where it is probably not intended.  I see two possible paths, either A) Require initialization to be performed async (don't use a static initializer), or B) Keep NetCore3.1 and lower using the WebRequest route and net5.0 and higher can utilize the sync method on HttpClient.  Any thoughts?

jon-armen:
Re: Item 3: Drop dependency for AWS SDK for .Net.

I'm also not seeing how this will be possible for several reasons:

1. Extensive use of AWS SDK's Internal Logger class. (Our own logger could be implemented to decouple us if we strongly desired to remove the sdk)
2. RegisterXRayForAllServices and XRayPipelineCustomizer utilize AmazonServiceClient to register services from the AWS SDK.  This functionality is, in my opinion a very desirable feature of this library.  See #241 for an alternative change that would strongly type the constraint for RegisterXRay<T> to enforce that only Amazon Service Client Interfaces are passed.

PR #134, opened 2020-05-29
Imdsv2 support

srprash:
*Issue #, if available:*

*Description of changes:*
Adding IMDSv2 support for fetching the EC2 instance metadata via the EC2Plugin. The metadata is fetched from [instance identity document](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) with a fallback to V1 if V2 fails.

### DIFF
--- a/sdk/src/Core/Plugins/EC2Plugin.cs
+++ b/sdk/src/Core/Plugins/EC2Plugin.cs
@@ -15,9 +15,12 @@
 // </copyright>
 //-----------------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
+using System.Net.Http;
+using System.Threading.Tasks;
 using Amazon.Runtime.Internal.Util;
-using Amazon.Util;
+using ThirdParty.LitJson;
 
 namespace Amazon.XRay.Recorder.Core.Plugins
 {
@@ -27,6 +30,7 @@ namespace Amazon.XRay.Recorder.Core.Plugins
     public class EC2Plugin : IPlugin
     {
         private static readonly Logger _logger = Logger.GetLogger(typeof(EC2Plugin));
+        private static readonly HttpClient _client = new HttpClient();
 
         /// <summary>
         /// Gets the name of the origin associated with this plugin.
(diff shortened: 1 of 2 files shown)


sdk/src/Core/Plugins/EC2Plugin.cs
```diff
@@ -60,36 +65,81 @@ public string ServiceName
...
+                string resp = DoRequest(metadata_base_url + "dynamic/instance-identity/document", HttpMethod.Get, header).Result;
+                dict = ParseMetadata(resp);
+            }
+            catch (Exception e)
             {
-                dict.Add("availability_zone", EC2InstanceMetadata.AvailabilityZone);
+                _logger.Error(e, "Exception occurred while getting EC2 metadata from IMDSv2");
+                _logger.DebugFormat("Unable to get metadata from IMDSv2 endpoint. Falling back to IMDSv1 endpoint");
+
+                // try IMDSv1 endpoint for metadata
```
anuraaga, line 91: I had a similar comment in the node PR - since there's no difference in the document endpoint for v2 and v1 I think we should just have error handling around the token request. The can continue with the same logic, including the token if it's available or otherwise try without it.

anuraaga: Here's my version in Java which takes that approach, the flow is just

- get token
- get document (include token if we got one)
- parse document

Don't need to treat v1 as a different endpoint

srprash: Oh yes. Good call out. The node PR discussion slipped out of my mind while writing this.

sdk/src/Core/Plugins/EC2Plugin.cs
```diff
@@ -60,36 +66,98 @@ public string ServiceName
         public bool TryGetRuntimeContext(out IDictionary<string, object> context)
         {
             context = null;
+            IDictionary<string, object> dict = new Dictionary<string, object>();
```
anuraaga, line 69: Instead of initialized to empty can we just declare the variable below?

sdk/src/Core/Plugins/EC2Plugin.cs
```diff
@@ -60,36 +66,98 @@ public string ServiceName
...
+                    headers.Add("X-aws-ec2-metadata-token", token);
+                }
+                string identity_doc_url = metadata_base_url + "dynamic/instance-identity/document";
+                string doc_string = DoRequest(identity_doc_url, HttpMethod.Get, headers).Result;
+                return ParseMetadata(doc_string);
             }
+            catch (Exception)
+            {
+                _logger.DebugFormat("Error occurred while getting EC2 metadata");
+                return null;
```
anuraaga, line 122: Not much experience with c# but is it possible to return an empty dictionary here instead of null? No points to do a null check at the caller if it's same semantics as empty

https://stackoverflow.com/a/54902869

srprash: Can't use `ImmutableDictionary` since it's [only supported](https://docs.microsoft.com/en-us/dotnet/api/system.collections.immutable.immutabledictionary-2.empty?view=netcore-3.1#moniker-applies-to) for .NetCore and .Net 5. Bummer :(

returning a new empty Dictionary<>() instead to avoid the null check.

normj:
Is there a reason you don't just use the `Amazon.Util.EC2InstanceMetadata` class to access the instance metadata? https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Util/_bcl%2Bnetstandard/EC2InstanceMetadata.cs

anuraaga:
Hi @normj thanks for the question! The AWS SDK is fairly heavyweight, with most of its logic in order to support SDK calls. As this is an instrumentation library for user apps, and many of those apps may not even use the AWS SDK in their code, we want to reduce the dependency footprint so that adding tracing has as little impact on the user's app as possible. If a small library containing only the metadata lookup was extracted out of `Core`, we would probably use that, but given the small amount of logic for the HTTP fetch and a simple parse, inlining seems ok for now. What do you think?

normj:
@anuraaga Totally agree you should not take a reference on AWSSDK.Core just for the metadata but it looks like the project already has a dependency on AWSSDK.Core

https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/AWSXRayRecorder.Core.csproj#L40-L43

anuraaga:
@normj Good eyes :) This and some of our other language SDKs have used it too for fetching sampling configuration, but because we proxy requests through the X-Ray domain, which signs the requests itself, it's actually overkill to pull in the machinery of the AWS SDK especially since we don't need to expose these models to users. So we've just begun to move away from using the SDK dependency hoping a future major version can be much more lightweight and improve user experience. Here's an [example](https://github.com/aws/aws-xray-sdk-java/pull/145) from the Java SDK, and in 3.0.0 we intend to remove the dependency.

@srprash Do you think we can add this point to #102? And maybe even add a comment in the `.csproj` as a pointer to people that find it.

normj:
@anuraaga that makes sense to move away from using the SDK for these small use cases and avoid bring in a large dependency that could possibly cause conflict with the versions the application is using.

srprash:
> @srprash Do you think we can add this point to #102? And maybe even add a comment in the `.csproj` as a pointer to people that find it.

@anuraaga Sure. That's a good point for visibility.

Issue #142, opened 2020-08-05
Hard to reproduce race condition in RulePoller and TargetPoller


jeffclary:
Found in Version: 2.7.2

I am seeing intermittent null reference exceptions when using XRay in an test environment, specifically running XUnit tests of an API (using Microsoft.AspNetCore.TestHost.TestServer) in a GitLab pipeline.

The TestServer is stopping with this stack trace:

```
The active test run was aborted. Reason: Test host process crashed : Unhandled Exception: System.NullReferenceException: Object reference not set to an instance of an object.
    at Amazon.XRay.Recorder.Core.Sampling.RulePoller.Start(Object state)
    at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
```

I believe this snippet from RulePoller.cs contains the code with the problem:

```
        internal void Poll(IConnector connector)
        {
            _connector = connector;
            _timer = InitializeTimer();
        }

        private Timer InitializeTimer()
        {
            return new Timer(Start, null, 0, 0);
        }

        internal async void Start(Object state)
        {
            try
            {
                await RefreshCache();
            }
            catch (Exception e)
            {
                _logger.Error(e, "Encountered an exception while polling sampling rules.");
            }
            finally
            {
                _timer.Change(GetDelay(), 0); // Next invocation with jitter
            }
        }

```
When `Poll` is first called, `_timer` is null.  It calls `InitializeTimer` and assigns the returned value to `_timer`.  But `InitializeTimer` creates a new `System.Threading.Timer` that calls `Start` with zero delay.  `Start` dereferences `_timer` in its finally block, which is where I think the null reference is happening.  Depending how the threads are scheduled, that finally block could be reached before the assignment to `_timer` happens.

I cloned this repo and tried to create a unit test to reproduce this issue, to no avail.  I suspect this race condition resolves differently on different architectures, container environments, etc.  Even in my build pipeline it does not fail every time.

I have not observed it as a failure, but there is very similar code that would subject to the same kind of race condition in [TargetPoller.cs](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/Sampling/TargetPoller.cs).

I am willing to fork and create a pull request for this issue, but since I cannot reproduce it at will, I won't be able to write a unit test that proves my fix.  I'd like some guidance from contributors on how to proceed and whether a pull request would be considered for merging.

srprash:
[...]

PR #154, opened 2020-10-07
Fixing race condition where the rule poller may start immediately before initializing the timer

srprash:
*Issue #, if available:* [142](https://github.com/aws/aws-xray-sdk-dotnet/issues/142)

*Description of changes:*
As mentioned in the issue thread, a better approach to start the rule poller is to initialize the `_timer` first with the call back and infinite due time, and then use `_timer.Change(0, 0)` to kick off the timer immediately.
This would ensure that the `_timer` instance is initialized before being accessed in the callback method itself.
Unfortunately, I couldn't come up with a unit test for this race condition. But I feel this refactoring is making the rule poller safer. 
Also, the issue mentioned that this problem could occur in the [TargetPoller](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/sdk/src/Core/Sampling/TargetPoller.cs#L55) as well, but its unlikely because the first invocation of timer happens after ~10s. So, I'm not touching that code in this PR.


yogiraj07 (approved):

lupengamzn:
LGTM

PR #157, opened 2020-10-13
Updated xraycontext to use a concurrent dictionary

### LINKED ISSUE #156: Exception - An item with the same key has already been added
Hi,

This is a copy of an issue I have logged over on the .net auto instrumentation repo.

https://github.com/aws/aws-xray-dotnet-agent/issues/19

However, I think the fix is to do with the SDK.

We are receiving the following exception intermittently.

```
Exception information: 
    Exception type: ArgumentException 
    Exception message: An item with the same key has already been added.
   at System.ThrowHelper.ThrowArgumentException(ExceptionResource resource)
   at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.Generic.Dictionary`2..ctor(IDictionary`2 dictionary, IEqualityComparer`1 comparer)
   at Amazon.XRay.Recorder.Core.AWSXRayRecorderImpl.AddRuleName(Segment newSegment, SamplingResponse sampleResponse)
   at Amazon.XRay.Recorder.Core.AWSXRayRecorderImpl.PopulateNewSegmentAttributes(Segment newSegment, SamplingResponse sampleResponse)
   at Amazon.XRay.Recorder.Core.AWSXRayRecorderImpl.BeginSegment(String name, String traceId, String parentId, SamplingResponse samplingResponse, Nullable`1 timestamp)
   at Amazon.XRay.Recorder.AutoInstrumentation.Utils.AspNetRequestUtil.ProcessHTTPRequest(Object sender, EventArgs e)
   at System.Web.HttpApplication.SyncEventExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute()
   at System.Web.HttpApplication.ExecuteStepImpl(IExecutionStep step)
   at System.Web.HttpApplication.ExecuteStep(IExecutionStep step, Boolean& completedSynchronously)
```

We receive this issue occasionally on application start. Once this issue occurs every request to the application then throws this exception. Causing the website to go down. I assume it could also happen on an application recycle.

It is hard to reproduce. I cannot reliably reproduce the issue, but it happens frequently. In our case when deploying an ElasticBeanstalk application version.

Our application makes lots of threaded http requests on startup. I believe this could be the cause.

I have been trying out this as a fix: https://github.com/sdolier/aws-xray-sdk-dotnet/commit/34167e89218d0357bc2bfa6f5821473664a46ab0

It is only a partial solution as there are other areas of the library that need to be updated to a ConcurrentDictionary.

I feel there should be global try/catch logic in this library, like in logging libraries (nlog, log4net). A monitoring/logging library should not be able to break a parent application.

sdolier:
To avoid concurrent thread access to the dictionary causing an exception

fixes #156 

### DIFF
--- a/sdk/src/Core/AWSXRayRecorderImpl.cs
+++ b/sdk/src/Core/AWSXRayRecorderImpl.cs
@@ -15,6 +15,7 @@
 // </copyright>
 //-----------------------------------------------------------------------------
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Reflection;
@@ -379,13 +380,12 @@ namespace Amazon.XRay.Recorder.Core
             IDictionary<string, string> xrayContext;
             if (newSegment.Aws.TryGetValue("xray", out object value))
             {
-                IDictionary<string, string> tempXrayContext = (Dictionary<string, string>) value;
-                xrayContext = new Dictionary<string, string>(tempXrayContext); // deep copy for thread safety
+                xrayContext = (ConcurrentDictionary<string, string>)value;
                 xrayContext[ruleNameKey] = ruleName;
             }
             else
             {
-                xrayContext = new Dictionary<string, string>();
+                xrayContext = new ConcurrentDictionary<string, string>();
                 xrayContext[ruleNameKey] = ruleName;
             }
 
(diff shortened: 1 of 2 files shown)


lupengamzn:
Thanks, will take a look later today

lupengamzn (approved):
LGTM

PR #159, opened 2020-11-04
Applied HttpClient to get sampling info

lupengamzn:
*Description of changes:*

1. Applied HttpWebRequest to get sampling rules and targets
2. Removed dependency on `AWSSDK.XRay` 

### DIFF
--- a/sdk/src/Core/Sampling/ServiceConnector.cs
+++ b/sdk/src/Core/Sampling/ServiceConnector.cs
@@ -14,13 +14,17 @@
 //      permissions and limitations under the License.
 // </copyright>
 //-----------------------------------------------------------------------------
+
 using System.Collections.Generic;
-using Amazon.XRay.Model;
 using System.Threading.Tasks;
 using System;
-using Amazon.Runtime;
 using Amazon.XRay.Recorder.Core.Internal.Utils;
 using Amazon.Runtime.Internal.Util;
+using Amazon.XRay.Recorder.Core.Sampling.Model;
+using System.Net;
+using ThirdParty.LitJson;
+using System.IO;
+using System.Text;
 
 namespace Amazon.XRay.Recorder.Core.Sampling
 {
@@ -32,19 +36,18 @@ namespace Amazon.XRay.Recorder.Core.Sampling
     class ServiceConnector : IConnector
     {
         private static readonly Logger _logger = Logger.GetLogger(typeof(ServiceConnector));
-        private AmazonXRayClient _xrayClient;
+        private XRayConfig _xrayConfig;
         private readonly object _xrayClientLock = new object();
         private const int Version = 1;
-        private readonly AmazonXRayConfig _config = new AmazonXRayConfig();
-        private readonly AWSCredentials _credentials = new AnonymousAWSCredentials(); // sends unsigned requests to daemon endpoint
         private readonly DaemonConfig _daemonConfig;
+        private readonly DateTime EpochStart = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
 
         /// <summary>
         /// Client id for the instance. Its 24 digit hex number.
         /// </summary>
         public string ClientID;
 
-        public ServiceConnector(DaemonConfig daemonConfig, AmazonXRayClient xrayClient) 
+        public ServiceConnector(DaemonConfig daemonConfig, XRayConfig xrayConfig) 
         {
             ClientID = ThreadSafeRandom.GenerateHexNumber(24);
             if (daemonConfig == null)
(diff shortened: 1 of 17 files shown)


sdk/src/Core/Sampling/ServiceConnector.cs
```diff
@@ -119,47 +164,101 @@ public async Task<GetSamplingRulesResponse>  GetSamplingRules()
...
+            samplingStatisticsModel.SamplingStatisticsDocuments = samplingStatisticsDocumentModels;
+
+            string requestContent = JsonMapper.ToJson(samplingStatisticsModel); // Marshall SamplingStatisticsDocument to json
+
+            Task<WebResponse> responseTask;
             lock (_xrayClientLock)
             {
                 RefreshEndPoint();
-                responseTask = _xrayClient.GetSamplingTargetsAsync(request);
+                responseTask = GetSamplingInfoAsync(_xrayConfig.ServiceURL + "/SamplingTargets", requestContent);
```
normj, line 178: Just getting the task here and doing the await outside of the lock defeats the purpose of the lock. The actual getting the response for the request will happen outside of the lock. So if the point of the lock was to not have multiple calls to the sevice url at the same time his won't work.

I assume this was done because you can't do an await with the lock keyword. Usually you get around that by directly using the .NET monitor class.

sdk/src/Core/Sampling/ServiceConnector.cs
```diff
@@ -53,61 +56,103 @@ public ServiceConnector(DaemonConfig daemonConfig, AmazonXRayClient xrayClient)
...
+
+            List<SamplingRule> samplingRules = UnmarshallSamplingRuleResponse(responseContent);
+
+            GetSamplingRulesResponse result = new GetSamplingRulesResponse(samplingRules);
+            return result;
+        }
+
+        private Task<WebResponse> GetSamplingInfoAsync(string url, string content = null)
+        {
+            var request = (HttpWebRequest)WebRequest.Create(url);
```
normj, line 108: In .NET Core using HttpWebRequest is highly frowned on as it is close to a deprecated class and has clean up issues. The preferred way is to use System.Net.Http.HttpClient.

sdk/src/Core/Sampling/ServiceConnector.cs
```diff
@@ -53,61 +56,103 @@ public ServiceConnector(DaemonConfig daemonConfig, AmazonXRayClient xrayClient)
...
-            List<SamplingRule> newRules = new List<SamplingRule>();
-            GetSamplingRulesRequest request = new GetSamplingRulesRequest();
-
-            Task<Model.GetSamplingRulesResponse> responseTask;
+            Task<WebResponse> responseTask;
             lock (_xrayClientLock)
             {
                 RefreshEndPoint();
-                responseTask = _xrayClient.GetSamplingRulesAsync(request);
+                responseTask = GetSamplingInfoAsync(_xrayConfig.ServiceURL + "/GetSamplingRules");
```
normj, line 94: Just getting the task here and doing the await outside of the lock defeats the purpose of the lock. The actual getting the response for the request will happen outside of the lock. So if the point of the lock was to not have multiple calls to the sevice url at the same time his won't work.

I assume this was done because you can't do an await with the lock keyword. Usually you get around that by directly using the .NET monitor class.

sdk/src/Core/Sampling/ServiceConnector.cs
lupengamzn: Right, fixed in the new commit

sdk/src/Core/Sampling/ServiceConnector.cs
lupengamzn: Switched to `HttpClient`

sdk/src/Core/Sampling/XRayConfig.cs
```diff
@@ -0,0 +1,28 @@
...
+//      or in the "license" file accompanying this file. This file is distributed
+//      on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+//      express or implied. See the License for the specific language governing
+//      permissions and limitations under the License.
+// </copyright>
+//-----------------------------------------------------------------------------
+
+namespace Amazon.XRay.Recorder.Core.Sampling
+{
+    public class XRayConfig
```
srprash, line 20: Need some docs for the class and the methods.

sdk/test/UnitTests/JsonMapperTest.cs
```diff
@@ -0,0 +1,275 @@
...
+
+        [TestMethod]
+        public void TestMarshallSamplingStatisticsDocumentsWithEmptyValues()
+        {
+            var samplingStatisticsModel = new SamplingStatisticsModel();
+            var samplingStatisticsDocumentModel = new SamplingStatisticsDocumentModel();
+
+            samplingStatisticsModel.SamplingStatisticsDocuments.Add(samplingStatisticsDocumentModel);
+
+            var expected = "{\"SamplingStatisticsDocuments\":[{\"ClientID\":null,\"RuleName\":null,\"RequestCount\":null,\"SampledCount\":null,\"BorrowCount\":null,\"Timestamp\":null}]}";
```
srprash, line 57: What's the response from the `GetSamplingTargets` API if the sampling statistics document has null/empty required fields? Do we need to handle any exceptions in that case?

srprash (commented):
Mostly looks good! Some doc related changes and a question. :)

sdk/test/UnitTests/JsonMapperTest.cs
lupengamzn: No, they didn't do sanity check for the fields. All of the sampling statistics document [fields](https://github.com/aws/aws-sdk-net/blob/7c88eeceb52869892adb1fe50c3f06c8f1a6b126/sdk/src/Services/XRay/Generated/Model/SamplingStatisticsDocument.cs#L38-L43) can be nullable and passing a sampling statistics document object with empty values to `GetSamplingTargets` API will not throw exception.

lupengamzn:
@srprash No, `ExpectContinue` is `Nullable`, its default value is `null`. When using `POST` method with content to get sampling targets in .NET framework, there will be an `Expect` header in request which will cause signature invalid error from X-Ray backend. We have to disable `ExpectContinue` header similar as what [AWS SDK](https://github.com/aws/aws-sdk-net/blob/master/sdk/src/Core/Amazon.Runtime/Internal/AmazonWebServiceRequest.cs#L41) does to avoid this issue.

srprash (approved):
Looks good! :)

lupengamzn:
Targeting .NET SDK [3.0.0](https://github.com/aws/aws-xray-sdk-dotnet/issues/102)

PR #169, opened 2020-12-28
Add support for EF Core 5

dashkan:
[IDbCommandInterceptor](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.diagnostics.idbcommandinterceptor?view=efcore-5.0) interface has changed in EF Core 5.0. All but one asynchronous methods now return [ValueTask](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=net-5.0). 

No exceptions are thrown when using the netstandard2.0 in a net50 / ef core 5 project. But EFInterceptor does not work.

This PR adds net5.0 as a target framework to the EntityFramework project and uses preprocessor directives to conditionally change the method signatures to support EF Core 5.

I can update PR to target netstandard2.1 in addition to OR instead of net50.

### DIFF
--- a/sdk/src/Handlers/EntityFramework/EFInterceptor.cs
+++ b/sdk/src/Handlers/EntityFramework/EFInterceptor.cs
@@ -80,7 +80,11 @@ namespace Amazon.XRay.Recorder.Handlers.EntityFramework
         /// <param name="result">Result from <see cref="IInterceptor"/>.</param>
         /// <param name="cancellationToken">Instance of <see cref="CancellationToken"/>.</param>
         /// <returns>Task representing the async operation.</returns>
+#if NET5_0
+        public override ValueTask<InterceptionResult<DbDataReader>> ReaderExecutingAsync(DbCommand command, CommandEventData eventData, InterceptionResult<DbDataReader> result, CancellationToken cancellationToken = default)
+#else
         public override Task<InterceptionResult<DbDataReader>> ReaderExecutingAsync(DbCommand command, CommandEventData eventData, InterceptionResult<DbDataReader> result, CancellationToken cancellationToken = default)
+#endif
         {
             ProcessBeginCommand(eventData);
             return base.ReaderExecutingAsync(command, eventData, result, cancellationToken);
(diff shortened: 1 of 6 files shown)


sdk/test/UnitTests/AWSXRayRecorder.UnitTests.csproj
```diff
@@ -2,64 +2,63 @@
 
   
   <PropertyGroup>
-    <TargetFrameworks>net452;netcoreapp2.0</TargetFrameworks>
+    <TargetFrameworks>net452;netcoreapp2.1;net5.0</TargetFrameworks>
```
lupengamzn, line 5: Any reason why you changed the framework from netcoreapp2.0 to 2.1?

sdk/src/Handlers/EntityFramework/EFInterceptor.cs
```diff
@@ -80,7 +80,11 @@ public override DbDataReader ReaderExecuted(DbCommand command, CommandExecutedEv
         /// <param name="result">Result from <see cref="IInterceptor"/>.</param>
         /// <param name="cancellationToken">Instance of <see cref="CancellationToken"/>.</param>
         /// <returns>Task representing the async operation.</returns>
+#if NET5_0
+        public override ValueTask<InterceptionResult<DbDataReader>> ReaderExecutingAsync(DbCommand command, CommandEventData eventData, InterceptionResult<DbDataReader> result, CancellationToken cancellationToken = default)
```
lupengamzn, line 84: Is it better to create a new interceptor class for .NET 5.0, similar as `EFInterceptor.net5.cs` ?

arturomonge:
I don't think adding NET 5.0 as a target is a good idea. EF 5.0 targets NET Standard 2.1, which means it is possible to use EF 5.0 in NET Core 3.1 too. So the target framework does not define which EF version is being used. I don't think we can make a single package that targets both EF versions conditionally.

We probably need to bump the package's version to 2.0 so v1.0 is for EF3.1, and v2.0 is for EF5.0. _Note: It would be ideal if we could match versions of the EF Handler package to the version of EF (v3.1 instead of v1.0 and v5.0 instead of v2.0)._

To do this, we could either:

1. Simply upgrade the existing project, and create a maintenance branch that would allow the v1.0 of the package to be built in case of a bug (but this means a separate workflow for maintenance), or
2. Add a second project to keep both in the master branch. This way, changes can be made to both projects independently in the same workflow.

btecu:
Looks like it needs a rebase.

dashkan:
@arturomonge I agree. I think aligning xray ef sdk w/ dotnet ef versioning is the best way to go. Sounds like team is moving towards using a diagnostic listener #174 anyways. I'll close this PR.

Issue #174, opened 2021-01-19
EntityFramework Handler does not work with EF5


arturomonge:
The current EntityFramework Handler does not work with the new EntityFramework 5.0. **No exceptions are raised and no indications are given of any problem**. The handler simply does not work. 

After some research, the problem is generated because the [IDbCommandInterceptor](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.diagnostics.idbcommandinterceptor?view=efcore-5.0) interface has changed in EF Core 5.0. Most asynchronous methods now return a [ValueTask](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=net-5.0).

arturomonge:
I've opened the following pull requests to address this problem in two possible ways:

https://github.com/aws/aws-xray-sdk-dotnet/pull/175 : Upgrades the existing EntityFramework NuGet package with EF5 support and bumps version.
https://github.com/aws/aws-xray-sdk-dotnet/pull/176 : Adds a new EntityFramework5 NuGet package with EF5 support.

There is also a pull request that was previously opened here (https://github.com/aws/aws-xray-sdk-dotnet/pull/169), which attempts to solve the problem by targeting NET5.0. A few questions have been raised about the solution in this PR, and that is why I created two additional PRs to offer two additional options to solve this issue.

lupengamzn:
Hey @arturomonge ,

Thank you for the contribution 

Currently we would like to keep the support for tracing Entity Framework in a single nuget package that customer can download to trace EF Core & EF 6. And implementing the [IDbCommandInterceptor](https://docs.microsoft.com/en-us/dotnet/api/microsoft.entityframeworkcore.diagnostics.idbcommandinterceptor?view=efcore-5.0) for .NET 5 will introduce a lot of platform complexity with the current interceptor as you have mentioned that supported .NET platforms are kind of overlapping with each other. Supporting .NET 5 is on our roadmap and we hope to come up with a better solution to address this, e.g., switching to [DiagnosticListener](https://docs.microsoft.com/en-us/ef/core/logging-events-diagnostics/diagnostic-listeners). Sorry for the inconvenience and please stay tuned!

arturomonge:
Understood, @lupengamzn. I'll close the pull requests and wait until you come up with a solution. Meanwhile, I created our own internal package specifically for EF5 using the modified sources, so we are not currently blocked by this.

[...]

PR #175, opened 2021-01-19
Upgrade the existing EntityFramework NuGet package with EF5 support

arturomonge:
This is one possible solution to https://github.com/aws/aws-xray-sdk-dotnet/issues/174

In this solution, we simply upgrade the existing package to EntityFramework 5, and bump the major version, as this is a breaking change. The new version of the package was also upgraded to NET Standard 2.1, as this is the minimum version of the standard supported by EF5.

In this solution, we would have to maintain a separate branch to provide updates to the EF3.1 version, which is not ideal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -388,11 +388,13 @@ using (var command = new TraceableSqlCommand("SELECT * FROM products", connectio
 
 AWS XRay SDK for .NET Core provides interceptor for tracing SQL query through Entity Framework Core (>=3.0).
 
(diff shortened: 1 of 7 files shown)


arturomonge:
As per the comments [here](https://github.com/aws/aws-xray-sdk-dotnet/issues/174#issuecomment-767182411), another solution is desired so I'm closing this pull request.

PR #176, opened 2021-01-19
Add new NuGet package of the EntityFramework Handler with support for EF5

arturomonge:
This is another possible solution to #174

In this solution, we add a new NuGet package named `AWSXRayRecorder.Handlers.EntityFramework5` with support for EF5. This way, both packages (the one with support for EF3.1 and the one with support for EF5) can be maintained separately in the same code line.


arturomonge:
As per the comments [here](https://github.com/aws/aws-xray-sdk-dotnet/issues/174#issuecomment-767182411), another solution is desired so I'm closing this pull request.

PR #179, opened 2021-02-11
Added release workflow

lupengamzn:
*Description of changes:*

Added workflow to release assets and generate tag on this repo


.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: 'v${{ github.event.inputs.version }}'
+          release_name: 'Release ${{ github.event.inputs.version }}'
+          body: 'Please refer [change-log](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/CHANGELOG.md) for more details'
+          draft: false
```
bhautikpip, line 26: we should set this flag as `true`

.github/workflows/release.yml
```diff
@@ -0,0 +1,27 @@
...
+      - name: Checkout master branch
+        uses: actions/checkout@v2
+
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: 'v${{ github.event.inputs.version }}'
```
bhautikpip, line 23: We should add capital `V` here? looks like all the .NET tag follows V2.9.0, V2.8.0 and so on from [here](https://github.com/aws/aws-xray-sdk-dotnet/tags)

.github/workflows/release.yml
lupengamzn: Updated

bhautikpip (approved):
LGTM

PR #181, opened 2021-02-16
Switching out AppVeyor for GH workflow

srprash:
Since we have tested GH workflows for continuous builds on our other SDKs and it has worked out well for us, it's probably a good idea that we make the switch here as well to keep it consistent. 


srprash:
Will need to remove the AppVeyor application from the repo as well.

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,30 @@
...
+                
+        steps:
+        - name: Checkout repository
+        - uses: actions/checkout@v2
+          
+        - name: Install dependencies
+          run: dotnet restore sdk/AWSXRayRecorder.sln --locked-mode
+          
+        - name: Build solution
+          run: dotnet build sdk/AWSXRayRecorder.sln --no-restore
```
lupengamzn, line 27: It's better for us to build package and test the dlls in `Release` configuration as it is the final stage that we [provided](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/buildtools/build.proj#L7) for end users

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,30 @@
...
+          
+        - name: Run tests
+          run: dotnet test sdk/test/UnitTests/bin/**/${{matrix.version}}/AWSXRayRecorder.UnitTests.dll
```
lupengamzn, line 30: Same here

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,30 @@
...
+            
+jobs:
+    build:
+        name: build
+        runs-on: ${{ matrix.os }}
+        
+        strategy:
+            fail-fast: false
+            matrix:
+                os: [windows-latest]
```
lupengamzn, line 16: Do we need unit test for .net core on `linux`  as well?

srprash: Yup. We should add linux as well. Good call.

.github/workflows/pr-build.yml
srprash: Yup

lupengamzn (approved):
LGTM

PR #183, opened 2021-02-18
Workflow fix

srprash:
*Description of changes:*
Noticed the workflow failure on [this](https://github.com/aws/aws-xray-sdk-dotnet/actions/runs/579560564) run because there was an `-` before `uses` key. Removing the `-` 

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -21,7 +21,7 @@ jobs:
                 
         steps:
         - name: Checkout repository
-        - uses: actions/checkout@v2
+          uses: actions/checkout@v2
           
         - name: Install dependencies
           run: dotnet restore sdk/AWSXRayRecorder.sln --locked-mode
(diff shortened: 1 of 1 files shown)


lupengamzn (changes requested):
The build failure on windows os seems like a similar issue discussed [here](https://github.com/actions/virtual-environments/issues/1090#issuecomment-748452120). Can you address it following the suggestions mentioned in the thread?

srprash:
Restore seems be looking into the local package cache by default. I think it will be good to define the package source as nuget.org as is done in [this PR](https://github.com/aws/aws-xray-sdk-dotnet/pull/180/files#diff-aab1525a907935fd3480702f8d2e6f430e949811da0b53c52f794cb96e94ed81) by adding a `Nuget.config` file. This way restore will download packages from nuget.org where it should find everything.

willarmiros:
@srprash is this PR still relevant? If not can we close it out?

srprash:
Closing the PR as it is not needed anymore.

PR #189, opened 2021-03-30
Add smoke test of distribution channel

lupengamzn:
*Issue #, if available:*

*Description of changes:*

Adds smoke test that periodically runs by building against https://www.nuget.org/. Doesn't publish metrics yet and will add later.


.github/workflows/continuous-monitoring.yml
```diff
@@ -0,0 +1,33 @@
+name: Continuous monitoring of distribution channels
+
+on:
+    workflow_dispatch:
+    schedule:
+      - cron: '0 */1 * * *'
```
wangzlei, line 6: Do we need canary every hour?

wangzlei (approved):

Issue #202, opened 2021-06-06
LitJson GUID type support


MadSciencist:
Hi.

Looks like the LitJson used for serialization does not properly handle GUID types. 

Minimum code to reproduce it:
```csharp
        public class AnythingWithGuid
        {
            public Guid Id { get; set; }
        }

        [TestMethod]
        public void SerializeObjectContainingGuid_ThrowsError()
        {
            var obj = new AnythingWithGuid();
            JsonMapper.ToJson(obj);
            Assert.IsNotNull(obj);
        }
```
The error: ThirdParty.LitJson.JsonException: Max allowed object depth reached while trying to export from type System.Guid.

Is there any workaround for it? Is such behavior documented?
Might be related: https://github.com/aws/aws-xray-sdk-dotnet/issues/109

Thanks,
Mateusz

srprash:
The [fix](https://github.com/aws/aws-xray-sdk-dotnet/pull/203) has been released in v2.11.0. Please let us know if this issue still persists.

PR #203, opened 2021-06-06
fix: ability to serialize GUIDs

MadSciencist:
*Issue https://github.com/aws/aws-xray-sdk-dotnet/issues/202*

*Added ability to serialize GUID.*


srprash (approved):
Thanks for adding this. It is unfortunate that we maintain this internal copy of LitJSON since the actual library is still not stable (current version is 0.17.0). I think that the lLitJSON library itself would be running into this issue. In that case, it would be good to bring to their notice. I'll try it out and open an issue with them. :)

willarmiros:
@srprash Can we merge this?

Issue #209, opened 2021-07-29
Adding the same exception to the xray trace handler in different subsegments throws a null reference exception


jon-armen:
This code should work, however currently throws a null reference exception because the existing cause does not have an exception associated.

```
            try
            {
                recorder.BeginSubsegment("child1");
                try
                {
                    try
                    {
                        recorder.BeginSubsegment("child2");
                        throw new AmazonServiceException();
                    }
                    catch (AmazonServiceException e)
                    {
                        recorder.AddException(e);
                        recorder.EndSubsegment();
                        throw;
                    }
                }
                catch (AmazonServiceException e)
                {
                    recorder.AddException(e);
                    recorder.EndSubsegment();
                    throw new EntityNotAvailableException("Dummy message", e);
                }
            }
            catch (EntityNotAvailableException e)
            {
                recorder.AddException(e);
                recorder.EndSegment();
            }
```

lupengamzn:
Hi @jon-armen ,

Thank you for bringing this up

This is indeed an edge case issue in X-Ray SDK while serializing duplicate exception. In the second `catch` statement, as the `AmazonServiceException` from previous step has already been recorded, SDK will only record the [Cause](https://github.com/aws/aws-xray-sdk-dotnet/blob/d2b64125ecdd734af220795d5b102bd96b037981/sdk/src/Core/Strategies/DefaultExceptionSerializationStrategy.cs#L156) to the exception descriptor. When the `AmazonServiceException` is wrapped as an inner exception and passed to next catch statement, the SDK will try to figure out if this inner exception has been recorded or not by checking the [Exception](https://github.com/aws/aws-xray-sdk-dotnet/blob/d2b64125ecdd734af220795d5b102bd96b037981/sdk/src/Core/Strategies/DefaultExceptionSerializationStrategy.cs#L192) from the descriptor, which, as I mentioned before, will not be recorded as only the `Cause` will be recorded for duplicate exception. This is why you see the null reference exception.

We'll work on this and get it fixed soon.

jon-armen:
Awesome, thank you @lupengamzn .  If it helps, I submitted #210 as a potential fix.

srprash:
Released in v2.11.0

PR #210, opened 2021-07-29
Finding the same exception in different subsegments should reuse the same cause id

jon-armen:
*Issue #, if available:*
#209

*Description of changes:*
Associate the existing descriptor's exception with the new descriptor so that it can be used in other subsegments.


jon-armen:
@lupengamzn , would you have any update as to when this could be incorporated, or if changes are required?

jon-armen:
@willarmiros , would you have any updates on when this could be reviewed?

PR #215, opened 2021-10-14
Add netcoreapp3.1 target framework

martincostello:
Add a target framework moniker for .NET Core 3.1.

## Context

We have a number of .NET Core 3.1 Lambda functions that use the X-Ray recorder, and due to the only Target Framework Moniker for .NET Core being `netstandard2.0`, it will publish the application with all of the relevant binary files, such as `Microsoft.AspNetCore.Http.dll`.

image

However, in .NET Core 3.0 and later, these binaries are no longer published as NuGet packages, and instead are included in the ASP.NET Core shared framework. This means that versions of the binary greater than [2.2.2](https://www.nuget.org/packages/Microsoft.AspNetCore.Http/2.2.2) are not available to be upgraded to explicitly, leading to "old" binaries that don't need to be used being published with the application.

As all versions of ASP.NET Core earlier than 3.1 are now out-of-support, any security vulnerabilities in these tools get flagged when using security scanning tools such as Nexus Sonatype.

image

As there are no "non-vulnerable" versions of the dependency available from NuGet (because they're in the shared framework), it is not possible to upgrade to a .NET Core 3.1 version of this dependency to resolve the issue.

This PR attempts to resolve this issue by adding a `netcoreapp3.1` Target Framework Moniker so that the relevant recorder libraries understand the ASP.NET Core Framework, and will not include the relevant binaries as part of the deployment package.

### DIFF
--- a/sdk/AWSXRayRecorder.nuspec
+++ b/sdk/AWSXRayRecorder.nuspec
@@ -30,6 +30,19 @@
         <dependency id="AWSXRayRecorder.Handlers.SqlServer" version="2.7.3" />
         <dependency id="AWSXRayRecorder.Handlers.System.Net" version="2.7.3" />
       </group>
+      <group targetFramework=".NETCoreApp3.1">
+        <dependency id="AWSXRayRecorder.Core" version="2.10.1" />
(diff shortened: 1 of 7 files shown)


lupengamzn (approved):
Thanks for contributing

PR #234, opened 2022-03-08
Removed blocking call which was producing deadlocks from EC2Plugin.

samunro:
The calls to `Task.Result` in `GetToken` and `GetMetadata` are blocking and can result in deadlocks if either the response or the timeout are handled on the same thread. See here for a similar example and explanation.

> "Don’t block on Tasks; use async all the way down."
https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html

One option here could have been to replace the calls to `Result` with the `async` and `await` keywords, but that would have required that modifications to the `IPlugin` interface and its consumers. That might be something that should be considered, but given that the call to `Result` was already making this code synchronous, nothing will be lost by using a synchronous API as an immediate fix.

There is a synchronous `Send` method which has been added to `HttpClient` in .Net 5 but it is not available in any of the targeted frameworks - .Net Framework 4.5, .Net Core 3.1 or .Net Standard 2.0.

 Although `HttpWebRequest` is not recommended for new development, it does offer a synchronous alternative and is the option that I have used in this pull request.

> "We don't recommend that you use `HttpWebRequest` for new development. Instead, use the [System.Net.Http.HttpClient](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netframework-4.5) class."
https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest?view=netframework-4.5

### DIFF
--- a/sdk/src/Core/Plugins/EC2Plugin.cs
+++ b/sdk/src/Core/Plugins/EC2Plugin.cs
@@ -17,9 +17,10 @@
 
 using System;
 using System.Collections.Generic;
-using System.Linq;
+using System.IO;
+using System.Net;
 using System.Net.Http;
-using System.Threading.Tasks;
+using System.Text;
 using Amazon.Runtime.Internal.Util;
 using ThirdParty.LitJson;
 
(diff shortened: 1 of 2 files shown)


jon-armen (approved):

srprash (approved):

srprash:
Hi @samunro 
Thanks for identifying this issue and providing the fix. We should switch to using `async` and `await` in our next major version bump.

samunro:
Thanks @srprash. Are there any plans to release a new NuGet package? The last one was in April last year.

https://www.nuget.org/packages/AWSXRayRecorder.Core/

PR #238, opened 2022-03-29
Warning cleanup

jon-armen:
*Description of changes:*

Cleaning up various warnings that were previously suppressed.

### DIFF
--- a/sdk/src/Core/AwsXrayRecorderBuilder.cs
+++ b/sdk/src/Core/AwsXrayRecorderBuilder.cs
@@ -182,10 +182,10 @@ namespace Amazon.XRay.Recorder.Core
         }
 
         /// <summary>
-        /// Adds the given plugin to builder
+        /// Sets the address for the xray daemon.
         /// </summary>
-        /// <param name="plugin">A specific plugin to add.</param>
-        /// <returns>The builder with plugin added.</returns>
+        /// <param name="address">The xray daemon address.</param>
+        /// <returns>The builder with the specified xray daemon address.</returns>
         public AWSXRayRecorderBuilder WithDaemonAddress(String address)
         {
             if (String.IsNullOrEmpty(address))
@@ -246,8 +246,8 @@ namespace Amazon.XRay.Recorder.Core
         /// <summary>
         /// Configures recorder with <see cref="ITraceContext"/> instance.
         /// </summary>
-        /// <param name="segmentEmitter">The provided <see cref="ISegmentEmitter"/> instance.</param>
-        /// <returns>The builder with ISegmentEmitter added.</returns>
+        /// <param name="traceContext">The provided <see cref="ITraceContext"/> instance.</param>
+        /// <returns>The builder with ITraceContext added.</returns>
         public AWSXRayRecorderBuilder WithTraceContext(ITraceContext traceContext)
         {
             _traceContext = traceContext ?? throw new ArgumentNullException("TraceContext");
@@ -258,8 +258,8 @@ namespace Amazon.XRay.Recorder.Core
         /// Configures recorder with provided <see cref="ExceptionSerializationStrategy"/>. While setting number consider max trace size
         /// limit : https://aws.amazon.com/xray/pricing/
         /// </summary>
-        /// <param name="size">stack frame size</param>
-        /// <returns>The builder with stack frame size added.</returns>
+        /// <param name="exceptionSerializationStartegy">An instance of <see cref="ExceptionSerializationStrategy"/></param>
+        /// <returns>The builder with exception serialization strategy added.</returns>
         public AWSXRayRecorderBuilder WithExceptionSerializationStrategy(ExceptionSerializationStrategy exceptionSerializationStartegy)
         {
             _exceptionSerializationStrategy = exceptionSerializationStartegy ?? throw new ArgumentNullException("ExceptionSerializationStartegy"); ;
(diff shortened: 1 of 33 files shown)


sdk/src/Core/AwsXrayRecorderBuilder.cs
```diff
@@ -182,10 +182,10 @@ public AWSXRayRecorderBuilder WithPlugin(IPlugin plugin)
         }
 
         /// <summary>
-        /// Adds the given plugin to builder
+        /// Sets the address for the xray daemon.
```
srprash, line 185: Wow. Never realized that the API and the doc were completely different.
Thanks so much for fixing these.

sdk/src/Core/ThirdParty/LitJson/IJsonWrapper.cs
```diff
@@ -1,3 +1,5 @@
+#pragma warning disable CS1587 // XML comment is not placed on a valid language element
```
srprash, line 1: Why does this warning needs to be suppressed? I don't see any XML comment on the namespace (or anywhere else in this file) that would raise this warning.

Wondering the same for other places in the `LitJson` files.

sdk/src/Handlers/EntityFramework/EFInterceptor.net45.cs
```diff
@@ -28,6 +28,10 @@ public class EFInterceptor : IDbCommandInterceptor
     {
         private readonly bool? _collectSqlQueriesOverride;
 
+        /// <summary>
+        ///
```
srprash, line 32:
```suggestion
        /// Initializes a new instance of the <see cref="EFInterceptor" /> class.
```

sdk/test/UnitTests/EfCoreTests.cs
```diff
@@ -86,6 +86,7 @@ public void Test_EFCore_unsuccessful_query()
         }
 
         [TestMethod]
+        [Obsolete]
```
srprash, line 89: Why is this test marked as obsolete?

sdk/src/Core/Properties/AssemblyInfo.cs
```diff
@@ -3,7 +3,7 @@
 using System.Runtime.CompilerServices;
 
 [assembly: ComVisible(false)]
-[assembly: CLSCompliant(true)]
+[assembly: CLSCompliant(false)]
```
srprash, line 6: Wondering if you have any ideas on making this code (and other assemblies) CLS compliant instead of setting it to `false`?

srprash (commented):
Thank you so much for doing this. :)

sdk/src/Core/ThirdParty/LitJson/IJsonWrapper.cs
jon-armen: It's the lack of the Amazon Copyright notice.  I'm assuming in this case Amazon can't claim notice for these files since they were pulled from somewhere else.

sdk/test/UnitTests/EfCoreTests.cs
jon-armen: context.Database.ExecuteSqlCommand is obsolete.  I've updated to context.Database.ExecuteSqlInterpolated, which is one of the recommended resolutions, and removed the obsolete marker.

sdk/src/Core/Properties/AssemblyInfo.cs
jon-armen: Usages of Microsoft.AspNetCore.Http.HttpRequest and Microsoft.Extensions.Configuration.IConfiguration are both not CLS compliant.  Given those exposed publicly, I don't believe we have a good path forward to indicate that these assemblies can be made to be CLS compliant.

sdk/src/Core/AwsXrayRecorderBuilder.cs
jon-armen: @srprash, no problems.  I'm sure you've got a lot going on with the XRay Team currently.  Are you able to give any updates yet on when we can expect a new release?  On a separate PR, I've pushed out a non-breaking form of support for net60.  It would be beneficial for a product I'm working on if we could get a release.  We're currently blocked from adopting net60 until we get the interceptor updates, and the current released assemblies reference unnecessary packages for net31 which contain security vulnerabilities.  Those issues have been fixed for months, but we haven't had a release to take advantage of those fixes.

sdk/src/Core/Properties/AssemblyInfo.cs
jon-armen: Taking a second look at it, I was able to apply CLSCompliant(false) to methods that were were in violation.  This leaves the assembly CLSCompliant, with small handful of methods and classes that are not.

sdk/src/Core/ThirdParty/LitJson/IJsonWrapper.cs
srprash: Right. This code was ported directly from LitJson so we can't put Amazon's Copyright notice.

sdk/src/Core/Properties/AssemblyInfo.cs
srprash: Yup. That's better. Thanks!

srprash (approved):

sdk/src/Core/AwsXrayRecorderBuilder.cs
srprash: @jon-armen I would love to get the new version released and have you (and others) unblocked. It's just that we are refactoring our build system and figuring out some technicalities on our end.
This is still a high priority for us and will get the new release out ASAP. Unfortunately, I can't disclose the ETA yet.

PR #253, opened 2022-06-23
Adding build-pack-release workflow

srprash:
The workflow has been tested in my fork with some test resources.
https://github.com/srprash/aws-xray-sdk-dotnet/actions/runs/2550725032


.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,127 @@
...
+
+        - name: Build solution
+          run: dotnet build .\sdk\AWSXRayRecorder.sln /p:Configuration=Release --no-restore /p:AssemblyOriginatorKeyFile=${{ github.workspace }}\awsxrayrecorder.dll.snk
+
+        - name: Assume signer role
+          run: |
+            $resp = aws sts assume-role --role-arn ${{ secrets.AWS_ARTIFACT_ACCESS_ROLE_ARN }} --external-id ${{ secrets.AWS_ROLE_EXTERNAL_ID }} --role-session-name AWSSigner-Session
+            $ak = echo $resp | jq -r ".Credentials.AccessKeyId"
+            $sk = echo $resp | jq -r ".Credentials.SecretAccessKey"
+            $tk = echo $resp | jq -r ".Credentials.SessionToken"
```
willarmiros, line 51: Didn't we use an action to assume roles before? Is this more/less secure? How can we be confident these creds will never be leaked?

srprash: The [`aws-actions/configure-aws-credentials`](https://github.com/aws-actions/configure-aws-credentials#assuming-a-role) action works only if the IAM role's trust policy allows GH OIDC to assume the role. It's not the case here, thus this following this way of assuming role.
I am able to mask the role's temporary credentials but still have concerns over the safety as [it doesn't seem to be reliable](https://github.com/actions/runner/issues/475). 
I'm working with the role owners for a safer way to assume roles in Github.

.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,96 @@
...
+          run: |
+            .\buildtools\sign_files.ps1 -Path .\sdk\src -Filter AWSXRayRecorder.*.dll -Recurse
+          env:
+            UNSIGNED_BUCKET: ${{ secrets.AWS_UNSIGNED_BUCKET_NAME }}
+            SIGNED_BUCKET: ${{ secrets.AWS_SIGNED_BUCKET_NAME }}
+
+        - name: Pack nugets
+          run: |
+            dotnet pack .\sdk\AWSXRayRecorder.sln  --no-build -c Release -o .\Deployment\nuget-packages
+            nuget pack .\sdk\AWSXRayRecorder.nuspec -OutputDirectory .\Deployment\nuget-packages -Exclude **
```
NathanielRN, line 61: Would it be worth it to split this into 2 steps to let them each fail separately?

Otherwise, mostly a nit.

```suggestion
        - name: Dotnet pack nugets
          run: >
            dotnet pack .\sdk\AWSXRayRecorder.sln 
              --no-build
              -c Release
              -o .\Deployment\nuget-packages
        - name: Nuget pack nugets
          run: >
            nuget pack .\sdk\AWSXRayRecorder.nuspec
              -OutputDirectory .\Deployment\nuget-packages
              -Exclude **
```

.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,96 @@
...
+        - name: Assume nuget role
+          uses: aws-actions/configure-aws-credentials@v1
+          with:
+            role-to-assume: ${{ secrets.NUGET_ACCESS_ROLE_ARN }}
+            aws-region: us-west-2
+
+        # TODO: uncomment this step once the rest of the workflow has been verified in prod.
+        #- name: Push packages to Nuget.org
+        #  run: |
+        #    $nugetKey = aws secretsmanager get-secret-value --secret-id ${{ secrets.NUGET_SECRETS_ID }} --region us-west-2 --output text --query SecretString
```
NathanielRN, line 78: Not sure if this is like bash, but if it was it would have to look like this right?

```suggestion
        #  run: |
        #    nugetKey="$(aws secretsmanager get-secret-value --secret-id ${{ secrets.NUGET_SECRETS_ID }} --region us-west-2 --output text --query SecretString)"
```

.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,96 @@
...
+          env:
+            UNSIGNED_BUCKET: ${{ secrets.AWS_UNSIGNED_BUCKET_NAME }}
+            SIGNED_BUCKET: ${{ secrets.AWS_SIGNED_BUCKET_NAME }}
+
+        - name: Pack nugets
+          run: |
+            dotnet pack .\sdk\AWSXRayRecorder.sln  --no-build -c Release -o .\Deployment\nuget-packages
+            nuget pack .\sdk\AWSXRayRecorder.nuspec -OutputDirectory .\Deployment\nuget-packages -Exclude **
+
+        - name: Upload nugets as actions artifact
```
NathanielRN, line 63: At first I was confused why we weren't then downloading the artifact, but I remember there was value to uploading it so we can see what was uploaded on the action itself.

nit:
```suggestion
        - name: Upload nugets to this GitHub Action run as an artifact
```

.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,96 @@
...
+            role-to-assume: ${{ secrets.NUGET_ACCESS_ROLE_ARN }}
+            aws-region: us-west-2
+
+        # TODO: uncomment this step once the rest of the workflow has been verified in prod.
+        #- name: Push packages to Nuget.org
+        #  run: |
+        #    $nugetKey = aws secretsmanager get-secret-value --secret-id ${{ secrets.NUGET_SECRETS_ID }} --region us-west-2 --output text --query SecretString
+        #    nuget push .\Deployment\nuget-packages\*.nupkg -Source https://api.nuget.org/v3/index.json -ApiKey $nugetKey
+            
+        # TODO: Chnage this step to directly create a public release once the workflow has been verified.
```
NathanielRN, line 81:
```suggestion
        # TODO: Change this step to directly create a public release once the workflow has been verified.
```

.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,96 @@
...
+        - name: Create draft release
+          id: create_release
+          uses: actions/create-release@v1
+          env:
+            GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
NathanielRN, line 86: Do you need to provide this? From the GitHub docs:

> Important: An action can access the GITHUB_TOKEN through the github.token context even if the workflow does not explicitly pass the GITHUB_TOKEN to the action.

However if we want to be explicit I'm okay with that. I can also see [the actions/create_release action says to do this](https://github.com/actions/create-release) in their documentation.

.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,96 @@
...
+          with:
+            tag_name: 'V${{ github.event.inputs.version }}'
+            release_name: 'Release ${{ github.event.inputs.version }}'
+            body: 'Please refer [change-log](https://github.com/aws/aws-xray-sdk-dotnet/blob/master/CHANGELOG.md) for more details'
+            draft: true
+            prerelease: false
+
+        - name: Clean up
+          run: |
+            rm .\awsxrayrecorder.dll.snk
```
NathanielRN, line 96: Can we clean this up earlier or request the key later to reduce the time we have it on the machine?

nit:
```suggestion
        - name: Clean up
          run: >
            rm .\awsxrayrecorder.dll.snk
```

buildtools/sign_files.ps1
```diff
@@ -0,0 +1,123 @@
...
+    Authenticode-sign a set of files.
+    You can pass either a file (or files e.g. -File file1, file2 - mind the limits of PS command line though)
+    or a folder with filter and recurse conditions (e.g. -Path folder -Filter *.dll,*.exe -Recurse).
+#>
+
+[CmdletBinding()]
+Param
+(
+    [Parameter(Mandatory=$true, ParameterSetName="File")]
+    [string[]]$File,
```
NathanielRN, line 14: Should this be called `Files` if its a array of strings of Files?

buildtools/sign_files.ps1
```diff
@@ -0,0 +1,123 @@
...
+
+    [Parameter(Mandatory=$true, ParameterSetName="Path")]
+    [string]$Path,
+
+    [Parameter(ParameterSetName="Path")]
+    [string[]]$Filter,
```
NathanielRN, line 20: Should this be called `Filters` to signify it supports multiple `Filters`?

.github/workflows/build-pack-release.yml
NathanielRN: Okay I just noticed in the `buildtools/sign_files.ps1` below this is how it is done in windows systems so I guess no change is needed here 🙂

buildtools/sign_files.ps1
```diff
@@ -0,0 +1,123 @@
+<#
+.Synopsis
+    Authenticode-sign a set of files.
```
NathanielRN, line 3:
```suggestion
    Authenticode-sign the set of provided files in-place.
```

buildtools/sign_files.ps1
```diff
@@ -0,0 +1,123 @@
...
+.DESCRIPTION
+    Authenticode-sign a set of files.
```
NathanielRN, line 5:
```suggestion
    Authenticode-sign the set of provided files in-place.
```

NathanielRN (commented):
This looks super great! Awesome work and I learned so much during my read through of it 🙂 Most of my comments are nits and hopefully helpful suggestions.

Let me know if it makes sense!

.github/workflows/build-pack-release.yml
NathanielRN: Sounds good. In that case is it still possible to separate arguments on a new line? According to [this StackOverflow post](https://stackoverflow.com/questions/605686/how-to-write-a-multiline-command) you can do it like this but I am not confident in it 😅

```suggestion
        - name: Pack nugets
          run: |
            dotnet pack .\sdk\AWSXRayRecorder.sln  ^
              --no-build ^
              -c Release ^
              -o .\Deployment\nuget-packages
            nuget pack .\sdk\AWSXRayRecorder.nuspec ^
              -OutputDirectory .\Deployment\nuget-packages ^
              -Exclude **
```

srprash: I'm not confident either. 😅 
As per this [yaml multiline doc](https://yaml-multiline.info/), "To get a newline using the folded style, leave a blank line by putting two newlines in.".
Then I think these command would look like this:
```yaml
- name: Pack nugets
   run: >
      dotnet pack
        .\sdk\AWSXRayRecorder.sln
        --no-build
        -c Release
        -o .\Deployment\nuget-packages

      nuget pack
        .\sdk\AWSXRayRecorder.nuspec
        -OutputDirectory .\Deployment\nuget-packages
        -Exclude **
```

.github/workflows/build-pack-release.yml
```diff
@@ -0,0 +1,113 @@
...
+            SIGNED_BUCKET: ${{ secrets.AWS_SIGNED_BUCKET_NAME }}
+          run: >
+            .\buildtools\sign_files.ps1
+            -Path .\sdk\src
+            -Filters AWSXRayRecorder.*.dll
+            -Recurse
+
+        - name: Pack nugets
+          run: >
+            dotnet pack
```
NathanielRN, line 73: Super cool 😮

NathanielRN (approved):
This looks excellent! Thanks for the great work! 😄

PR #255, opened 2022-07-01
Improvement on signing step of release workflow + uncomment push to nuget.org step

srprash:
*Description of changes:*
- Invoking the signing script separately for each project path. This avoids signing unnecessary files from `\obj\` folder as well, as seen in [this test run](https://github.com/aws/aws-xray-sdk-dotnet/runs/7139466034?check_suite_focus=true).
  We only need to sign `.dll` files from the '\bin\` path because that's what ultimately gets shipped. We don't need to spend signing resources on intermediary files.

- Uncomment the step to publish the packages to nuget.org

### DIFF
--- a/.github/workflows/build-pack-release.yml
+++ b/.github/workflows/build-pack-release.yml
@@ -62,6 +62,7 @@ jobs:
           env:
             UNSIGNED_BUCKET: ${{ secrets.AWS_UNSIGNED_BUCKET_NAME }}
             SIGNED_BUCKET: ${{ secrets.AWS_SIGNED_BUCKET_NAME }}
+          #TODO: There is probably a better way to pass in a list of paths as a single parameter to the script.
           run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-pack-release.yml
```diff
@@ -93,13 +96,19 @@ jobs:
...
+            --region us-west-2
+            --output text
+            --query SecretString
+
+            nuget push
+            .\Deployment\nuget-packages\*.nupkg
+            -Source https://api.nuget.org/v3/index.json
+            -ApiKey $nugetKey
             
-        # TODO: Change this step to directly create a public release once the workflow has been verified.
```
srprash, line 102: @willarmiros I know we talked earlier to make this step create an actual release. But I think we can keep this as draft just to be able to edit the release if needed. For example, the 3.x release will need to be a pre-release, and so on.

NathanielRN (approved):
Awesome! 😄

Issue #267, opened 2022-10-21
X-Ray Cannot Find a Handler of Type `EndpointResolver`


chrisoverzero:
Since the release of AWSSDK service client libraries which includes AWSSDK.DynamoDBv2 version 3.7.100, Amazon.XRay.Recorder.Handlers.AwsSdk can no longer customize service clients due to being unable to find a handler for the type `EndpointResolver`. The service clients fail to be constructed, as a consequence.

Here is the stack trace for attempting to resolve an instance of `IAmazonDynamoDB`:

```json
{
  "errorType": "InvalidOperationException",
  "errorMessage": "Cannot find a handler of type EndpointResolver",
  "stackTrace": [
    "at Amazon.Runtime.Internal.RuntimePipeline.AddHandlerAfter[T](IPipelineHandler handler)",
    "at Amazon.XRay.Recorder.Handlers.AwsSdk.Internal.XRayPipelineCustomizer.Customize(Type serviceClientType, RuntimePipeline pipeline) in /_/sdk/src/Handlers/AwsSdk/Internal/XRayPipelineHandler.cs:line 748",
    "at Amazon.Runtime.Internal.RuntimePipelineCustomizerRegistry.ApplyCustomizations(Type type, RuntimePipeline pipeline)",
    "at Amazon.Runtime.AmazonServiceClient.BuildRuntimePipeline()",
    "at Amazon.Runtime.AmazonServiceClient..ctor(AWSCredentials credentials, ClientConfig config)",
    "at Amazon.DynamoDBv2.AmazonDynamoDBClient..ctor(AWSCredentials credentials, AmazonDynamoDBConfig clientConfig)"
  ]
}
```

normj:
X-Ray team this was a change that went out with the AWS .NET SDK yesterday. Sorry about the miss here updating X-Ray. I'll send over a PR to address the change shortly.

chrisoverzero:
Ah, I wasn't quite certain where to file the report. Apologies if this was off the mark.

normj:
@chrisoverzero The code change has to go here so this was the right spot.

chrisoverzero:
That's great, thanks so much.

Lookin' forward to Amazon.XRay.Recorder.Handlers.AwsSdk 2.9.2.

roelvandenbrand:
Testing with the current versions (core 3.7.100.7 & xray 2.11.1) and it does not work.

srprash:
[...]

PR #268, opened 2022-10-21
Fix issue #267 using version 3.7.100 of AWS SDK causing exception not finding EndpointResolver

normj:
*Issue #, if available:*
https://github.com/aws/aws-xray-sdk-dotnet/issues/267

*Description of changes:*
Starting with version 3.7.100 of the AWS .NET SDK the `EndpointResolver` is replaced with a service specific version of the handler. This breaks the X-Ray handler because X-Ray was inserting itself just after `EndpointResolver` in the SDK's request pipeline.

The fix is for X-Ray to inject itself before the `RetryHandler` which happens after endpoint resolutions. So the position will be the same in the request pipeline.

I have confirmed this fix works correctly by deploying a Lambda function and seeing the SDK traces in the console. I tested both with 3.7.100 and a version 3.7.10.1 to confirm we are not breaking users that haven't upgraded to the latest .NET SDK.


normj:
Looking into the test failure

klementi (approved):

boblodgett (approved):

philasmar (approved):

srprash (approved):

Issue #277, opened 2022-11-23
EC2Plugin not working


a-jackson:
Using v2.13.0 I'm not getting EC2 metadata in my traces.

I enabled logging and get "Failed to get token for IMDSv2", "Error occurred while getting EC2 metadata" and "Could not get instance metadata."

Debugging finds that the exception is "System.ArgumentException: '' is not a supported encoding name. For information on defining a custom encoding, see the documentation for the Encoding.RegisterProvider method. (Parameter 'name')" which comes from [here](https://github.com/aws/aws-xray-sdk-dotnet/blob/2d0e1f87c713567f047c2262d13a70a9415ac012/sdk/src/Core/Plugins/EC2Plugin.cs#L145). Seems there is no `response.ContentEncoding` set.

PR #278, opened 2022-11-23
ec2plugin: fallback to utf8 if response encoding not set

a-jackson:
*Issue #, if available:* #277

*Description of changes:*

If `repsonse.ContentEncoding` is not set then this will fallback to UTF8 to prevent an exception.

### DIFF
--- a/sdk/src/Core/Plugins/EC2Plugin.cs
+++ b/sdk/src/Core/Plugins/EC2Plugin.cs
@@ -142,7 +142,9 @@ namespace Amazon.XRay.Recorder.Core.Plugins
                     throw new Exception("Unable to complete the request successfully");
                 }
 
-                var encoding = Encoding.GetEncoding(response.ContentEncoding);
+                var encoding = !string.IsNullOrEmpty(response.ContentEncoding)
+                    ? Encoding.GetEncoding(response.ContentEncoding)
+                    : Encoding.UTF8;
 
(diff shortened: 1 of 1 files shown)


srprash:
Hi @a-jackson 
Thanks for opening this PR. Can you add a corresponding unit test for the fix too?

a-jackson:
Hi @srprash 

Just managed to get back to this, there doesn't appear to be any unit tests for the EC2Plugin at all right now and I'm not really sure how I'd do it given that is sends Http requests to the hard coded metadata endpoint.

srprash:
Hi @a-jackson 
The unit tests for EC2 plugin are here. The tests use a mocked `EC2Plugin` to return some predefined values. Can you give the tests a look and see if you can incorporate the new unit test for this change without a lot of effort?

a-jackson:
I'm not really sure what you're after with that? If it's using a mocked `EC2Plugin` then it isn't going to be executing the code I have changed.

srprash:
Ah I see. We mock the `DoRequest` method from the plugin so it will be difficult to test the change without having to re-structure the existing unit tests. 
Given that this is a very minor change to add more code robustness and the unit testing effort is comparatively large, I think it is acceptable to merge this PR as is.

Thanks @a-jackson for contributing!

srprash (approved):

